PR #1, opened 2018-10-26
convert to data frame in toolAggregate()

bs538:
toolAggregate() fails with a somewhat unhelpful message when called with a tibble (instead of a data frame) as a relation matrix. tibbles are the default return for some newer read-in function (such as readxl:read_excel ), so I thought it was worth adding the option to use them by adding a conversion to data frame at the beginning of toolAggregate().

Minimal (non-)working example:

```
library(madrat)
library(magclass)
library(tibble)
x <- new.magpie(cells_and_regions = c('A', 'B', 'C', 'D'),
                years = 2000,
                names = 'dummy',
                fill = 1:4)
relmatrix <- tibble(CountryCode = c('A', 'B', 'C', 'D'),
                    RegionCode = c('XX', 'XX', 'YY', 'YY'))
# this works
toolAggregate(x, rel = data.frame(relmatrix))
# this fails (prior to fix)
toolAggregate(x, rel = relmatrix)
```
Edit: formatting of example

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Package: madrat
 Type: Package
 Title: May All Data be Reproducible and Transparent (MADRaT) *
-Version: 1.53.0
-Date: 2018-10-18
+Version: 1.53.1
+Date: 2018-10-26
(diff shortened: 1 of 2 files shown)


tscheypidi:
thanks!

PR #6, opened 2019-11-15
added debug setting and improved cache file handling

tscheypidi:
- added debug setting which will force try=TRUE on all calculations
- malformed cache files will no longer interrupt calculations, instead cache file will be ignored


bodirsky:
will make a testrun now.

PR #11, opened 2020-02-25
add tool function to average over time using toolAggregate

k4rst3ns:
**including metadata updates** 
that maybe have to be check, since both toolAggregate and toolTimeAverage will have a stamp in the metadata like this: 
$calcHistory

 levelName
1 toolTimeAverage(x, averaging_range = 8) 
2  °--toolAggregate(x, rel = mat, dim = 2)


k4rst3ns:
* added bugfix to proper averaging
* rewrite the test for annual time set completeness
* add a non annual option (without any checks)

PR #24, opened 2020-07-03
Added option of 67420 cells in toolTimeSpline

FelicitasBeier:
toolTimeSpline did not work for case of 67420 cells. Added if statement for this special case

### DIFF
--- a/R/toolTimeSpline.R
+++ b/R/toolTimeSpline.R
@@ -6,7 +6,7 @@
 #' @param dof degrees of freedom per 100 years (similiar to an average range), is a proxy for the smoothness of the spline
 #' 
 #' @return approximated data in magclass format
-#' @author Kristine Karstens
+#' @author Kristine Karstens, Felicitas Beier
 #'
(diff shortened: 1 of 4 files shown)


R/toolTimeSpline.R
```diff
@@ -41,12 +41,17 @@ toolTimeSpline <- function(x, dof=NULL){
...
   x <- as.array(x)
-  
+
   # fill in data with spline approximations/interpolations
   tmpspline <- function(x,dof) return(smooth.spline(x,df=dof, control.spar=list(high=2))$y)
   out <- apply(x, c(1,3) ,tmpspline , dof=dof)
   dimnames(out)[[1]] <- getYears(x)
-  out <- as.magpie(out)
+  
+  if (length(out[1,,1])==67420){
```
tscheypidi, line 50: there shouldn't be a special case for data with 67420 cells. Just run as.magpie(out, spatial=2,temporal=1) in all instances.

PR #25, opened 2020-07-03
Add option to aggregate data to an additional regional resolution ("R…

dklein-pik:
…EG+GLO+EXTRA").

### DIFF
--- a/R/calcOutput.R
+++ b/R/calcOutput.R
@@ -97,9 +97,17 @@ calcOutput <- function(type,aggregate=TRUE,file=NULL,years=NULL,round=NULL,suppl
     # rename aggregate arguments from old to new convention, if necessary
     if(toupper(aggregate)=="GLO") aggregate <- "global"
     if(toupper(gsub("+","",aggregate,fixed = TRUE))=="REGGLO") aggregate <- "region+global"
-    if(!all(strsplit(aggregate,"+",fixed=TRUE)[[1]] %in% rel_names)) {
-      stop("Illegal setting aggregate = ",aggregate,"! Make sure that all arguments 
-            which should be passed to the specific calc function are given with its name (e.g. arg=BLA)")
+    
+    # Ignore columns in 'aggregate' that are not defined in one of the mappings. 
+    # Stop if 'aggregate' contains none of the columns defined in one of the mappings.
+    aggregate_splitted <- strsplit(aggregate,"+",fixed=TRUE)[[1]]
+    common_columns <- aggregate_splitted %in% rel_names
+    if(all(!common_columns)) {
+      stop("None of the columns given in aggregate = ",aggregate," could be found in the mappings!")
+    } else {
+      if(any(!common_columns)) vcat(verbosity = 0,'Omitting ',aggregate_splitted[!common_columns],' from aggregate = ',aggregate,' because it does not exists in the mappings.')
+      # Use those columns only for aggregation that exist in either of the mappings
+      aggregate <- paste0(aggregate_splitted[common_columns],collapse = "+")
     }
   }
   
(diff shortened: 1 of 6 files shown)


R/calcOutput.R
```diff
@@ -89,6 +89,7 @@ calcOutput <- function(type,aggregate=TRUE,file=NULL,years=NULL,round=NULL,suppl
     # rename column names from old to new convention, if necessary
     if(any(names(rel[[r]])=="CountryCode")) names(rel[[r]])[names(rel[[r]])=="CountryCode"] <- "country"
     if(any(names(rel[[r]])=="RegionCode")) names(rel[[r]])[names(rel[[r]])=="RegionCode"] <- "region"
+    if(any(names(rel[[r]])=="ExtraRegions")) names(rel[[r]])[names(rel[[r]])=="ExtraRegions"] <- "extra"
```
tscheypidi, line 92: The lines with countryCode and RegionCode are only needed for backwards compatibility. As the ExtraRegions approach is new I would like you ask you to remove this line and useinstead the new terminology from the beginning

R/calcOutput.R
```diff
@@ -7,7 +7,7 @@
 #' @param type output type, e.g. "TauTotal". A list of all available source
 #' types can be retrieved with function \code{\link{getCalculations}}.
 #' @param aggregate Boolean indicating whether output data aggregation should be performed or not, "GLO" (or "glo") for aggregation to one global region,
-#' "REG+GLO" (or "regglo") for a combination of regional and global data.
+#' "REG+GLO" (or "regglo") for a combination of regional and global data, "REG+GLO+EXTRA" (or "reggloextra") same as "REG+GLO", but with an additional regional aggregation.
```
tscheypidi, line 10: It might be worthwhile to update the param description, but instead of adding another special case it would make more sense to explain the general possibility to have other aggregations and that you just have to mention the corresponding column name here. Please also do not introduce this alternative terminology which I had to introduce for backwards compatibility. It should be just the name of the column and "+" will allow to merge aggregation rules of more than one column

R/calcOutput.R
```diff
@@ -97,6 +98,7 @@ calcOutput <- function(type,aggregate=TRUE,file=NULL,years=NULL,round=NULL,suppl
     # rename aggregate arguments from old to new convention, if necessary
     if(toupper(aggregate)=="GLO") aggregate <- "global"
     if(toupper(gsub("+","",aggregate,fixed = TRUE))=="REGGLO") aggregate <- "region+global"
+    if(toupper(gsub("+","",aggregate,fixed = TRUE))=="REGGLOEXTRA") aggregate <- "region+global+extra"
```
tscheypidi, line 101: please do not introduce this alternative syntax. For REGGLO it is just in there to be backwards compatible.

R/calcOutput.R
dklein-pik: A comment in the code that this is only for backwards compatibility could have helped.

R/calcOutput.R
dklein-pik: Since you created the new convention and know it better, could you add the explanation please? Could you include a hint if the user has to provide ALL column names of the mapping, i.e. "RegionsCode" and "GLO" if the mapping contains these columns?

R/calcOutput.R
dklein-pik: See above.

R/calcOutput.R
dklein-pik: How would you have to call calcCoutput if your mapping has the columns:  "X | CountryCode | RegionCode | meins"?

tscheypidi: you mean if you want to aggregate by all three? You would write aggregate = "country+region+meins"
Ideally, new mappings should have the columns X|country|region|meins, but your example would work as well due to the backwards compatibility fix in the code for CountryCode and RegionCode

tscheypidi: or if your data is on country level and you want to aggregate to region and meins it would be aggregate="region+meins"

dklein-pik: Thanks!

PR #26, opened 2020-07-31
enable regionscode for mappings which contain more than 3 columns

LaviniaBaumstark:


tscheypidi:
In this PR it is problematic that you always delete the first column in strict mode. Deletion only makes sense if the first column is just a naming column for the mapping! In the previous version this could be detected just via the number of columns (as that mapping was only allowed to consist of two columns). If you allow for more columns you need a different methodology to determine  whether the first column is a naming column or not!

In addition: The documentation has not been updated according to you changes and still claims that strict=TRUE only works for 2 or 3 column data.

PR #28, opened 2020-08-24
Added rds mapping files support

tscheypidi:
- added rds support in toolGetMapping, toolConvertMapping, toolAggregate and calcOutput

- toolAggregate can read now all file formats supported by toolGetMapping

- tests which need an internet connection are now skipped if no connection is available


k4rst3ns (approved):
Seems fine to me.

PR #30, opened 2020-09-14
Added custom class support in calcOutput

tscheypidi:
- added custom class support via x$class argument
- added corresponding tests


giannou (approved):
"..will not be available.." would be the correct documentation entry (currently reads "..will not be able.."

PR #34, opened 2020-11-19
Expand and automatize the search for mapping files

giannou:
Without this change, extra mappings from other packages are not found by `madrat`. Not sure however if people are supposed to also add extra mappings directly from the `getConfig()$mappingfolder`


giannou:
TODOS:
- change the way the "active package" is sought for in `calcOutput` (use `prepFunctionName`)
- remove `for` loop
- change `toolGetMapping` to allow for vectors in the `where` argument. The order in which mappings will be looked for has to be: absolute path, `getConfig("mappingfolder"), active package, other loaded packages

tests/testthat/test-toolGetMapping.R
```diff
@@ -6,8 +6,12 @@ test_that("toolGetMapping works", {
                              CountryCode = c("ABW", "AFG", "AGO","AIA", "ALA", "ALB"), 
                              RegionCode = c("LAM", "OAS", "SSA", "LAM","EUR", "NEU")), 
                         row.names = c(NA, 6L), class = "data.frame")
+  write.csv(expected,file = "expected.csv",quote = FALSE,row.names = FALSE)
```
tscheypidi, line 9: CRAN policy requires that these files which are written, must be stored in a temporary directory

tscheypidi (changes requested):
please make sure that tests only create/change/delete files in temporary directories (CRAN policy). Otherwise, the PR is good to go.

tests/testthat/test-toolGetMapping.R
giannou: not sure how to deal with this issue.. is this file written in a temporary directory? https://github.com/pik-piam/madrat/blob/9a41765c7cc595b6a98070ed0ce5cc7fa4ba775b/tests/testthat/test-calcOutput.R#L133

tscheypidi: yes, it is. You can see that a few lines above (mainfolder = tempdir())
in your case you can solve it easily by using tempdir() or tempfile() to use a temporary directory where you put the file or directly a temporary file (if the file name does not matter)

giannou:
converted to draft while running the pre-processing of magpie and remind once more for testing

tscheypidi (approved):

PR #35, opened 2020-11-20
tweaked metadata in downloadSource

tscheypidi:


dklein-pik (commented):
Looks good! Could you give some advise to the user what to put into "settings" and what to put into the unnamed list that is returned directly? Could we move up the unnamed list a bit, otherwise the user might miss that he/she has to also provide elements at the bottom of the function?

PR #39, opened 2021-02-12
Performance tweak in toolAggregate

tscheypidi:
Making use of the Matrix package toolAggregate is now significantly faster in various situations


giannou:
Nice! Has this been tested with both fullYXZ functions?

tscheypidi:
has been tested with fullValidation and fullMAgPIE

giannou:
Should we run a test with fullREMIND also?

tscheypidi:
from my perspective this is not needed as there is quite some overlap between the tested setup and fullREMIND...in addition toolAggregate comes which a rather good suite of tests which covered all emerging problems

giannou:
> from my perspective this is not needed as there is quite some overlap between the tested setup and fullREMIND...in addition toolAggregate comes which a rather good suite of tests which covered all emerging problems

fair enough

giannou (approved):

PR #40, opened 2021-02-16
Added functions toolSplitSubtype and metadataGFZ

tscheypidi:
- added metadataGFZ
- added toolSplitSubtype
- fixed problem in downloadSource and readSource that both functions were trying to work with potentially illegal folder names


R/toolSplitSubtype.R
```diff
@@ -0,0 +1,32 @@
+#' Tool: SplitSubtype
+#' 
+#' This function can split a subtype string into smaller entities based on 
+#' a given separator and check the these entities exist in a reference list
```
FelicitasBeier, line 4: "check the these entities", probably: "check if or whether these entities exist"

FelicitasBeier (approved):

PR #48, opened 2021-04-23
Improve madrat documentation

pfuehrlich-pik:
This PR
- fixes some typos in the vignette
- replaces MADRaT with madrat
- fixes a bug where the default argument of downloadTau was not taken into account
- improves some error messages


R/toolAggregate.R
```diff
@@ -120,7 +120,15 @@ toolAggregate <- function(x, rel, weight=NULL, from=NULL, to=NULL, dim=1, wdim=N
         if (length(from) == 0) {
           maxMatchColumn <- which.max(sapply(lapply(rel,intersect,items),length))
           unmappedItems <- setdiff(items, rel[[maxMatchColumn]])
-          stop(paste("The following items were not found in the mapping:", toString(unmappedItems)))
+          missingItems <- setdiff(rel[[maxMatchColumn]], items)
+
+          if (length(unmappedItems) > 0) {
+            warning(paste("The following items were not found in the mapping file:", toString(unmappedItems)))
```
tscheypidi, line 126: I would not call it "mapping file", just "mapping"

R/readSource.R
```diff
@@ -118,7 +118,11 @@ readSource <- function(type,subtype=NULL,convert=TRUE) {
   if(source_missing) {
     # does a routine exist to download the source data?
     if(type %in% getSources(type="download")) {
-      downloadSource(type, subtype=subtype)
+      if (is.null(subtype)) {
+        downloadSource(type)
+      } else {
+        downloadSource(type, subtype=subtype)
+      }
```
tscheypidi, line 125: Is that distinction necessary? Isn't the default setting of subtype in downloadSource not NULL anyway?

R/toolstartmessage.R
```diff
@@ -15,12 +15,18 @@
...
 #' \dontrun{
-#' tmp <- function(bla=NULL) {
-#'   startinfo <- toolstartmessage("+")
-#'   print(bla)
-#'   toolendmessage(startinfo,"-")
-#'   }
-#' tmp(bla=99)
+#' innerFunction <- function() {
+#'   startinfo <- madrat:::toolstartmessage("+")
+#'   print("inner")
```
tscheypidi, line 20: use cat or vcat instead of print, e.g. `vcat(1, "inner")`

R/toolstartmessage.R
```diff
@@ -15,12 +15,18 @@
 #' @examples
 #' 
 #' \dontrun{
```
tscheypidi, line 17: I believe in they way you have written the example it should work out of the box and you can remove the `\dontrun` statement

R/vcat.R
```diff
@@ -75,8 +75,8 @@ vcat <- function(verbosity,...,level=NULL, fill=TRUE, show_prefix=TRUE) {
   
   if (!is.null(level)) {
     if (level == "+") {
-      # add empty space
-      options(gdt_nestinglevel = paste0(">", getOption("gdt_nestinglevel")))
+      # indent output using the ~ character (findBottleneck expects ~ at line start)
+      options(gdt_nestinglevel = paste0("~", getOption("gdt_nestinglevel")))
```
tscheypidi, line 79: perhaps we could set the indentation symbol via our config

tscheypidi (changes requested):
I put some minor remarks / requests. In addition to them please run `lucode2::buildLibrary` for an additional check prior to submission, additional housekeeping and for increasing version number and date

R/toolAggregate.R
pfuehrlich-pik: In Zeile 107 steht "Cannot find given region mapping file!", ich änder das mal überall.

pfuehrlich-pik: Hm nein, dort geht es tatsächlich um ein file. Also mach ich die Änderung doch nur an dieser einen Stelle.

R/readSource.R
pfuehrlich-pik: You're right, it's not necessary here.

tscheypidi (approved):

PR #49, opened 2021-04-28
Hashing based on sample content and file size

pfuehrlich-pik:
fingerprintFiles now uses the first 300 bytes of file content and the file's size for creating a fingerprint


tscheypidi (approved):
looks good

PR #50, opened 2021-05-04
bugfixes in cacheArgumentsHash

tscheypidi:
- generated cache was depending on locale (as ordering of lower and upper case arguments is handled differently)
- comparison of arguments was not working properly


pfuehrlich-pik (approved):
Looks good, although I would prefer using method="radix"

PR #51, opened 2021-05-04
Optional filename tag for retrieveData

pfuehrlich-pik:
I applied some auto-formatting, so if actual changes are hard to see only look at commits after the initial formatting.
fullXYZ functions may now `return(list(tag = "mytag"))`, which will result in a filename like rev0_h12_test_mytag.tgz (with mytag as suffix)

### DIFF
--- a/R/retrieveData.R
+++ b/R/retrieveData.R
@@ -97,9 +97,20 @@ retrieveData <- function(model, rev = 0, dev = "", cachetype = "rev", ...) {
     "rev", rev, dev, "_", regionscode, "_", argsHash, tolower(model),
     ifelse(getConfig("debug") == TRUE, "_debug", "")
   )
-  sourcefolder <- paste0(getConfig("outputfolder"), "/", collectionname)
-  if (!file.exists(paste0(sourcefolder, ".tgz")) || getConfig("debug") == TRUE) {
+
+  .characterInCollectionname <- function(regexCharacter) {
+    return(grepl(regexCharacter, collectionname, fixed = TRUE))
+  }
+  if (any(unlist(lapply(strsplit("[](){}*+?|^$.\\", "")[[1]], .characterInCollectionname)))) {
+    warning(paste0("At least one of the regex characters [](){}*+?|^$.\\ appeared in collectionname",
+                   " - this might lead to unexpected caching behavior"))
+  }
+
+  if (length(dir(path = getConfig("outputfolder"),
+                 pattern = paste0(collectionname, ".*\\.tgz"))) == 0
+      || getConfig("debug") == TRUE) {
     # data not yet ready and has to be prepared first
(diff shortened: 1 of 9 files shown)


k4rst3ns:
I am unsure, is there still a hash in the tgz-file name summarizing the settings of the full*-function call?
so that something like rev0_h12_test_mytag_<hash_including_all_parameters_of_full*_call>.tgz
I would be happy to get an info about what the name includes at the moment and what the changes introduces at the example of the fullCELLULARMAGPIE from mrmagpie.

pfuehrlich-pik:
Yes the argument hash is still there. It is omitted if no arguments are passed to the fullXYZ method, that's why it is missing in my example. The file name is the same as before, here is the code where it is assembled:
```
collectionname <- paste0(
    "rev", rev, dev, "_", regionscode, "_", argsHash, tolower(model),
    ifelse(getConfig("debug") == TRUE, "_debug", "")
  )
```
With this PR a tag is appended to the end of the filename (but before .tgz) if the fullXYZ method ends with `return(list(tag="mytag"))`. Here's the corresponding code:
```
    collectionname <- paste0(collectionname,
                             ifelse(is.list(returnedValue) && exists("tag", returnedValue),
                                    paste0("_", returnedValue$tag),
                                    ""))
```
So we have a filename like rev0_h12_test_1234567_mytag.tgz, where 1234567 is the argsHash.
For fullCELLULARMAGPIE nothing will change, unless it returns list(tag = "anything").

pfuehrlich-pik:
The tag is not used in the cachefolder name. However, caching is not applied to fullXYZ functions, but rather to the calc functions called from fullXYZ. These are cached as before, the tag does not show up here, but the cache files are still valid when running the fullXYZ again with a different tag.

k4rst3ns (approved):
From user side no issue just praise for these changes.

R/retrieveData.R
```diff
@@ -1,132 +1,175 @@
...
+  collectionname <- paste0(
+    "rev", rev, dev, "_", regionscode, "_", argsHash, tolower(model),
+    ifelse(getConfig("debug") == TRUE, "_debug", "")
+  )
+
+  .characterInCollectionname <- function(regexCharacter) {
+    return(grepl(regexCharacter, collectionname, fixed = TRUE))
+  }
+  if (any(unlist(lapply(strsplit("[](){}*+?|^$.\\", "")[[1]], .characterInCollectionname)))) {
+    warning(paste0("At least one of the regex characters [](){}*+?|^$.\\ appeared in collectionname",
```
tscheypidi, line 105: can you add a test which triggers this new warning?

tscheypidi (changes requested):

tscheypidi (approved):

PR #55, opened 2021-05-10
Logging evaluated args

pfuehrlich-pik:
With this PR the actual values passed to a madrat function are logged, inste of unevaluated expressions (variable names). Here's a comparison example:

variable <- "paper"
readSource("Tau", subtype = variable)

log without this PR:
Run readSource("Tau", subtype = variable)

log with this PR:
Run readSource("Tau", subtype = "paper")


tscheypidi (approved):
looks good

PR #58, opened 2021-05-14
Same hashes on different systems

pfuehrlich-pik:
With this PR the hashes calculated on different systems should be the same. This enables a workflow of creating cache files on one system and using them on another, e.g. via sshfs.

`method = "radix"` is used for `sort` and `order`, because it is locale independent.


tscheypidi (approved):
looks good

PR #61, opened 2021-05-18
tryCatch fingerprint

### LINKED ISSUE #60: Missing error handling on packageDescription()$Version crashes everything
https://github.com/pik-piam/madrat/blob/df126b80de02d19ff52ea3726c370307fbd8c42a/R/prepFunctionName.R#L82
https://github.com/pik-piam/madrat/blob/df126b80de02d19ff52ea3726c370307fbd8c42a/R/calcOutput.R#L269

When packages happen to be updated while their `DESCRIPTION` file is being read, `packageDescription()` returns `NA`, which obviously does not have a `Version` field, resulting in `$ operator is invalid for atomic vectors`.  
See [here](https://mattermost.pik-potsdam.de/rd3/pl/4xbneqcqopgwzmkksndsujyymr), [here](https://mattermost.pik-potsdam.de/rd3/pl/fnjrz9no97rpig1gfb66qree9o), [here](https://mattermost.pik-potsdam.de/rd3/pl/wctf8nid6bygugm81chsnmomeo), and [here](https://mattermost.pik-potsdam.de/rd3/pl/aq9xpt3tetyd7ykgkpusdjjgry).

-----

Also, this setup reads the version of the package in the library directory, which isn't necessarily the same as the one attached in the R session, corrupting the cache.

pfuehrlich-pik:
Wrap fingerprint in tryCatch, so calculation can continue if fingerprinting is broken.
Also fixes #60

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '206854336'
+ValidationKey: '206884125'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,6 +1,6 @@
 {
   "title": "madrat: May All Data be Reproducible and Transparent (MADRaT) *",
-  "version": "1.102.4",
+  "version": "1.102.5",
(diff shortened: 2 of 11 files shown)


tscheypidi (approved):
good to go

PR #62, opened 2021-05-18
use fingerprintError files if forcecache=TRUE 

pfuehrlich-pik:


k4rst3ns (approved):
Seems fine for me, but I am not into the details of the functions.
From a user perspective, this fix makes sense as I try to ignore the fingerprint when using forcecache=TRUE, and so also fingerprint errors should be ignored.

tscheypidi (approved):

PR #64, opened 2021-05-18
Bugfix sort encoding

### LINKED ISSUE #59: df126b8 breaks REMIND input data preparation
[log-19638235.log](https://github.com/pik-piam/madrat/files/6494858/log-19638235.log)

pfuehrlich-pik:
Non-utf8 strings no longer lead to crashes (fixes #59) and sorting is now locale independent everywhere.
Added the internal robustOrder and robustSort functions, which are wrapper functions reound base::order and base::sort, but they convert character vectors to utf8 and always use local independent radix sort. In madrat these functions should be used instead of base::order and base::sort.

### DIFF
--- a/R/robustOrder.R
+++ b/R/robustOrder.R
@@ -1,0 +1,33 @@
+#' robustOrder, robustSort
+#' 
+#' robustOrder: A wrapper around base::order that always uses the locale independent method = "radix". If the
+#' argument x is a character vector it is converted to utf8 first.
+#' robustSort: A convenience function using order to sort a vector using radix sort. The resulting vector will have the
+#' same encoding as the input although internally character vectors are converted to utf8 before ordering.
+#' 
+#' @param ... One or more vectors of the same length
+#' @param na.last If TRUE missing values are put last, if FALSE they are put first, if NA they are removed
+#' @param decreasing If TRUE decreasing/descending order, if FALSE increasing/ascending order. For the "radix" method,
+#' this can be a vector of length equal to the number of arguments in ... . For the other methods, it must be length one.
+#' @param method Default is "radix", which is locale independent. The alternatives "auto" and "shell" should not be used
+#' in madrat because they are locale dependent.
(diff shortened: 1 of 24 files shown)


R/robustOrder.R
```diff
@@ -0,0 +1,33 @@
...
+#' @param decreasing If TRUE decreasing/descending order, if FALSE increasing/ascending order. For the "radix" method,
+#' this can be a vector of length equal to the number of arguments in ... . For the other methods, it must be length one.
+#' @param method Default is "radix", which is locale independent. The alternatives "auto" and "shell" should not be used
+#' in madrat because they are locale dependent.
+#' @seealso \code{\link[base]{order}}
+#' @author Pascal Führlich
+robustOrder <- function(..., na.last = TRUE, decreasing = FALSE, method = "radix") {
+  args <- list(...)
+  if (any(lapply(args, length) == 0)) {
+    return()
```
tscheypidi, line 19: I don't get the meaning of that part. The function should have no return value if any of the arguments provided has the length 0?

R/robustOrder.R
```diff
@@ -0,0 +1,33 @@
...
+  }
+  args <- lapply(args, sapply, function(element) {
+    ifelse(is.character(element), enc2utf8(element), element)
```
tscheypidi, line 22: I've read that the `ifelse` function is rather slow. I would suggest to change it to:
```suggestion
    if(is.character(element)) element <- enc2utf8(element)
    return(element)
```

tscheypidi (changes requested):
looks good, just one suggestion and one question for clarification

R/robustOrder.R
pfuehrlich-pik: This check prevents an error when an empty vector is sorted. `return()` still returns `NULL`, which works, but to better match the behavior of base::order I changed it to return vector(mode = "integer", length = 0).

PR #65, opened 2021-05-19
bugfixing incorrect/missing argumentHash if fingerprint hashing failed

tscheypidi:
bugfixing incorrect/missing argumentHash if fingerprint hashing failed but forcecache was active


k4rst3ns (approved):
Seems fine to me.

PR #67, opened 2021-05-20
fix "cannot xtfrm data frames"

pfuehrlich-pik:


pfuehrlich-pik:
I always try to do the formatting/linting in a different commit. In this case the first 2 commits contain the actual changes. Basically I just replaced
```
tmp <- tmp[robustOrder(tmp[2], tmp[1]), ]
```
with
```
tmp <- tmp[robustOrder(tmp[[2]], tmp[[1]]), ]
```
(single [] vs. double [[]])

tscheypidi:
I should have checked the single commits...thanks for the hint.

tscheypidi (approved):

PR #68, opened 2021-05-21
added Cache in fingerprintFiles

tscheypidi:
- added hashCache in fingerprintFiles which should improve performance for folders with many files
- added option to ignore a folder for fingerprint via creating a file `DONTFINGERPRINT` 

Note: fingerprints which contain files/folders will change due to a different structure of the object hashed by fingerprintFiles


R/fingerprint.R
```diff
@@ -30,76 +30,135 @@
...
+      files$size <- file.size(files$path)
+      # create key to identify entries which require recalculation
+      files$key <- apply(files[c("name", "mtime", "size")], 1,
+        digest,
+        algo = getConfig("hash")
+      )
+    }
+
+    getHashCacheName <- function(path) {
+      # return file name for fileHash cahe if the given path belongs to a source folder,
```
pfuehrlich-pik, line 128:
```suggestion
      # return file name for fileHash cache if the given path belongs to a source folder,
```

pfuehrlich-pik (approved):
Looks good to me!

PR #76, opened 2021-08-17
added try environment for aggregation statement

tscheypidi:


pfuehrlich-pik (approved):
I prefer handling errors with `tryCatch` instead of `try`. Checking the class attribute with `"try-error" %in%` is quite ugly in my opinion. Hadley Wickham also suggests using `tryCatch` instead of `try`: https://adv-r.hadley.nz/conditions.html#ignoring-conditions
However, in this case the `tryCatch` approach is a little problematic, because you want to return from calcOutput.
It's also super confusing and dangerous that there is a `try` variable, but also the `try` function at the same time, but since that's a function argument we probably have to stick with it.

PR #79, opened 2021-09-23
add .local argument to setConfig

pfuehrlich-pik:
`madrat::setConfig(..., .local = TRUE)` uses `withr::local_options` instead of `options` to only change the configuration temporarily.


tscheypidi (approved):
looks good

PR #81, opened 2021-10-15
Fix fingerprinterror

pfuehrlich-pik:
If the hash cache file was completely empty a fingerprintError happened, and the potentially corrupt cache file was read without notice. This PR fixes that.

### DIFF
--- a/R/cacheName.R
+++ b/R/cacheName.R
@@ -22,7 +22,7 @@
 #' Sources/Calculations should be returned
 #' @param globalenv	Boolean deciding whether sources/calculations in the global
 #' environment should be included or not
-#' @return cached data, if cache is available, otherwise NULL
+#' @return Name of fitting cache file, if available, otherwise NULL
 #' @author Jan Philipp Dietrich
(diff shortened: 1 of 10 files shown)


R/cacheName.R
```diff
@@ -36,7 +36,7 @@ cacheName <- function(prefix, type, args = NULL,  graph = NULL, mode = "put", pa
 
   fp <- fingerprint(name = paste0(fpprefix, type), graph = graph, details = (mode == "put"),
                     packages = packages, globalenv = globalenv)
-  if (identical(fp, "fingerprintError") && !isTRUE(getConfig("forcecache"))) {
+  if (fp == "fingerprintError" && !isTRUE(getConfig("forcecache"))) {
```
tscheypidi, line 39: This will fail if `fp` is `NULL`. one could perhaps use `as.character` to drop the attributes:

```suggestion
  if (identical(as.character(fp), "fingerprintError") && !isTRUE(getConfig("forcecache"))) {
```

PR #86, opened 2021-11-04
fix broken log messages, code cleanup

pfuehrlich-pik:
- fix broken log messages when running madrat functions with `::` instead of loading madrat via `library`
- fix crash during logging when running madrat function via do.call
- replace setwd with local_dir
- use `setConfig(..., .local = TRUE)` where possible
- replace deprecated getRegions with getItems
- replace `1:x` with `seq_len(x)`/`seq_along(x)`
- lots of code style improvements

These changes were tested by running the cellularmagpie rev 4.63 and remind preprocessings.


R/readSource.R
```diff
@@ -146,6 +146,6 @@ readSource <- function(type, subtype = NULL, convert = TRUE) {
   x <- .getData(type, subtype, prefix)
   on.exit(toolendmessage(startinfo, "-"))
   x <- clean_magpie(x)
-  setwd(cwd)
+  local_dir(cwd)
```
tscheypidi, line 149: This is probably not needed here, right?

R/toolendmessage.R
```diff
@@ -11,20 +11,6 @@
 #' hierarchy level by 1).
 #' @author Jan Philipp Dietrich
 #' @seealso \code{\link{toolstartmessage}}, \code{\link{vcat}}
-#' @examples
```
tscheypidi, line 14: why did you remove the example?

R/toolendmessage.R
```diff
@@ -1,7 +1,7 @@
 #' Tool: End message
 #'
-#' Function writes a process end message and performs some diagnostics
-#'
+#' This function writes a process end message and performs some diagnostics. It is always called after a corresponding
+#' call to toolstartmessage.
```
tscheypidi, line 4:
```suggestion
#' call to \code{\link{toolstartmessage}}.
```

tscheypidi (changes requested):
looks good, only a few minor comments/modification requests. In addition to the comments in the code: Can you update `setConfig` so that it indicates in its log messages of local modification that it is running with .local = TRUE (e.g. "Local configuration update" instead of "Configuration update") as otherwise it does not become clear that settings will autmatically switch back to their previous defaults?

R/getMainfolder.R
```diff
@@ -1,61 +1,68 @@
...
-  # ask for folder to use
-  if(interactive() || .testmode){
-    repeat {
-      s <- tolower(readline("madrat mainfolder for data storage not set! Do you want to set it now? (y/n) "))
-      if(.testmode) s <- "y"
-      if(s %in% c("y","n")) break
+  if (!is.na(mainfolder)) return(mainfolder)
+
+  questionLoop <- function(question, testModeAnswer) {
+    if (.testmode) {
```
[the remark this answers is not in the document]
pfuehrlich-pik: What's the advantage of doing it that way over the way it's implemented now?

R/readSource.R
pfuehrlich-pik: I have no idea why, but the remind preprocessing broke because of this... I used git bisect and some trial and error to identify exactly this as the reason, but I still wonder why.

R/toolendmessage.R
pfuehrlich-pik: It was exactly the same as the example in toolstartmessage I believe.

tscheypidi: I see

R/getMainfolder.R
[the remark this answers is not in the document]
tscheypidi: R allows to directly access variables in parent environments, but this is dangerous as the function declaration creates the believe that the function only depends on `question` and `testModeAnswer` where in fact it also requires to have `.testmode` available. Therefor, I always try to avoid these situations and always try to put all variables required by a function into its list of arguments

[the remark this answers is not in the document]
pfuehrlich-pik: I agree that pure functions are the most robust kind of functions, but I strongly disagree with calling this dangerous. I'd say this is maybe a 2/10 danger score, where global variables aka options are at least a 7/10 and NSE is 9/10, and we are using that all over the place in madrat. In fact I would even argue that in this case an additional .testmode variable in the inner function is more confusing, because then you have two different variables with the same name which could have different values. I'd say this kind of shadowing is more dangerous than just using the outer .testmode in the inner function. Also, if I see an inner function I don't expect it to be independent of the parent scope, because one of the main reasons to use inner functions is precisely that they can use the outer variables. In lapply calls it is also very common to create inner functions which use variables from the enclosing scope.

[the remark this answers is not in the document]
pfuehrlich-pik: Are you okay with the new default argument solution I just pushed?

R/readSource.R
pfuehrlich-pik: Okay, I took a deeper look into this. `withr` does not work together with `on.exit`. `withr` adds the code to be executed after the function to an attribute of the current environment and sets an on.exit handler that reads that attribute and runs the code. In this case `on.exit` overwrote (is that a word?) the on.exit handler with its own code. Setting `local_dir` at the end had an effect, because it was setting the `withr` on.exit handler again, so that the cleanup code of the first `local_dir` was re-enabled.

pfuehrlich-pik: tldr: Always use `withr::defer` instead of `on.exit` when using any `withr` functions.

pfuehrlich-pik: What do you think of deprecating `on.exit` in our linter settings and suggesting `withr::defer` as an alternative?

PR #93, opened 2021-12-02
wait for running download in readSource

pfuehrlich-pik:
This PR deals with the following situation: A source is being downloaded (downloadSource) and in another parallel running R session that source is read (readSource). At the moment readSource would try to read the partially downloaded data and probably crash. With this PR readSource will recognize the running download and wait. downloadSource now writes into a folder called `<type>-downloadInProgress` and when the download is finished that folder is renamed to `<type>`.

### DIFF
--- a/R/downloadSource.R
+++ b/R/downloadSource.R
@@ -3,16 +3,13 @@
 #' Download a source. The function is a wrapper for specific functions designed
 #' for the different possible source types.
 #'
-#'
-#' @param type source type, e.g. "IEA". A list of all available source types
-#' can be retrieved with function \code{\link{getSources}("download")}.
-#' @param subtype For some sources there are subtypes of the source, for these
-#' source the subtype can be specified with this argument. If a source does not
-#' have subtypes, subtypes should not be set.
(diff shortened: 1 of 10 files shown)


R/downloadSource.R
```diff
@@ -73,48 +73,50 @@ downloadSource <- function(type, subtype = NULL, overwrite = FALSE) {
...
-           "\" does already exist! Delete folder or activate overwrite to proceed!")
+      stop('Source folder for source "', typesubtype,
+           '" does already exist! Delete folder or activate overwrite to proceed!')
     }
   }
 
-  dir.create(typesubtype, recursive = TRUE)
-  absolutePathTypesubtype <- normalizePath(typesubtype)
-  local_dir(absolutePathTypesubtype)
+  downloadInProgressDirectory <- paste0(typesubtype, "-downloadInProgress")
```
tscheypidi, line 81: wouldn't the current setup mean that parallel preprocessings would now start to download the same source data in parallel into the same folder?

tscheypidi (approved):
looks good

PR #97, opened 2022-02-04
Add madrat bundles

tscheypidi:
- Added the possibility to create non-aggregated data bundles
- added `bundleCompile` function
- integration bundle creation/compilation process into `retrieveData` workflow
- added tests


R/retrieveData.R
```diff
@@ -26,166 +27,230 @@
...
 
-  regionmapping <- getConfig("regionmapping")
-  if (!file.exists(regionmapping)) {
-    regionmapping <- toolGetMapping(regionmapping, type = "regional", returnPathOnly = TRUE)
   }
-  regionscode <- regionscode(regionmapping, label = uselabels)
 
-  rev <- numeric_version(rev)
+  # data not yet ready and has to be prepared first
+  sourcefolder <- paste0(getConfig("outputfolder"), "/", cfg$collectionName)
```
pfuehrlich-pik, line 80:
```suggestion
  sourcefolder <- file.path(getConfig("outputfolder"), cfg$collectionName)
```

R/retrieveData.R
```diff
@@ -26,166 +27,230 @@
...
+      dir.create(dirname(mappath), recursive = TRUE)
     }
+    file.copy(regionmapping, mappath)
+  }
+  # copy mapping to output folder
+  try(file.copy(regionmapping, sourcefolder, overwrite = TRUE))
+  try(saveRDS(list(package = attr(cfg$functionName, "package"),
+                   args = argumentValues[!(names(argumentValues) %in% names(cfg$setConfig))],
+                   bundleArguments = cfg$bundleArguments, sessionInfo = sessionInfo()),
+              file.path(sourcefolder, "config.rds"), version = 2))
```
pfuehrlich-pik, line 101: Maybe a warning in case of a failure in these try statements would be a good idea here?

R/retrieveData.R
```diff
@@ -26,166 +27,230 @@
...
-    with_dir(sourcefolder, {
-      returnedValue <- do.call(functiononly, args)
-    })
-    if ("tag" %in% names(returnedValue)) {
-      if (grepl(pattern = "debug", returnedValue$tag)) {
-        warning("The tag returned in a fullXYZ function should not include the word 'debug'")
+  }
+
+  with_dir(sourcefolder, {
+    suppressWarnings(tar(paste0("../", cfg$collectionName, ".tgz"), compression = "gzip"))
```
pfuehrlich-pik, line 169:
```suggestion
    suppressWarnings(tar(file.path("..", paste0(cfg$collectionName, ".tgz")), compression = "gzip"))
```

R/retrieveData.R
```diff
@@ -26,166 +27,230 @@
...
+  if (!all(tmp)) {
+    stop("Unknown argument(s) \"", paste(names(cfg$input)[!tmp], collapse = "\", \""), "\" in retrieveData")
+  }
+
+  # are arguments used in both - setConfig and the fullFunction? -> warning
+  tmp <- intersect(formalArgs(setConfig), names(cfg$fullDefault))
+  if (length(tmp) > 0) {
+    warning(
+      "Overlapping arguments between setConfig and retrieve function (\"",
+      paste(tmp, collapse = "\", \""), "\")")
```
pfuehrlich-pik, line 215:
```suggestion
      paste(tmp, collapse = '", "'), '")')
```

R/fullEXAMPLE.R
```diff
@@ -33,5 +33,6 @@ fullEXAMPLE <- function(rev = 0, dev = "", extra = "Example argument") {
     message("Here you could execute code for a hypothetical development version called \"test\"")
   }
   # return is not required, but can be used to append a tag to the resulting filename
-  return(list(tag = "customizable_tag"))
+  return(list(tag = "customizable_tag",
+              pucTag = "tag"))
```
pfuehrlich-pik, line 37: What is the pucTag doing?

R/setConfig.R
```diff
@@ -133,6 +133,10 @@ setConfig <- function(regionmapping = NULL, # nolint
   firstsetting <- TRUE
   info <- NULL
 
+  if (!is.null(regionmapping)) {
+    if (file.exists(regionmapping)) regionmapping <- normalizePath(regionmapping)
+  }
```
pfuehrlich-pik, line 138:
```suggestion
  if (!is.null(regionmapping) && file.exists(regionmapping)) {
    regionmapping <- normalizePath(regionmapping)
  }
```

pfuehrlich-pik (approved):

R/fullEXAMPLE.R
tscheypidi: same as `tag` but for put files...I add some information about it to the documentation of `retrieveData`

R/setConfig.R
tscheypidi: woops, not sure why I wrote it that way

R/retrieveData.R
pfuehrlich-pik: I realized try will still print the error to stdout/via `cat`, but I don't think that's good enough, because these messages are so easy to miss in a long log. Also, it's confusing to have a completely normal message starting with "Error: ...". If instead the error was converted to a warning both of these problems are solved, because by default warnings are printed at the end of the log (easy to find) and they are printed as "Warning: ..." instead of starting with "Error: ...". Furthermore our "warnings are not okay" policy should apply here which is not entirely clear if we would have a normal cat message.
tldr: Can we please use errorToWarning instead of try here? (And in all similar situations in this PR and in the future?)
```r
errorToWarning <- function(expr) {
  maybeError <- try(expr, silent = TRUE)
  if (inherits(maybeError, "try-error")) {
    warning(maybeError)
  }
}
```

R/setConfig.R
```diff
@@ -37,9 +37,9 @@
 #' @param outputfolder The folder all outputs should be written to. In the
 #' default case this argument is set to NA meaning that the default folder
 #' should be used which is <mainfolder>/output
-#' @param bundlefolder The folder all output bundle should be written to. In the
-#' default case this argument is set to NA meaning that the default folder
-#' should be used which is <mainfolder>/bundle
+#' @param pucfolder The folder all portable unaggregated collections (puc) files
+#' should be written to. In the default case this argument is set to NA meaning
+#' that the default folder should be used which is <mainfolder>/puc
```
pfuehrlich-pik, line 42: They are not only written to that location, but also read from there, right?
```suggestion
#' @param pucfolder The path where portable unaggregated collection (puc) files
#' are located. NA by default, which means <mainfolder>/puc
```

pfuehrlich-pik (changes requested):
Looks good, the only thing important to me is turning errors to warnings in situations where they are just converted to normal cat messages via `try` at the moment.

vignettes/madrat-puc.Rmd
```diff
@@ -0,0 +1,96 @@
...
+  %\VignetteIndexEntry{Portable unagreggated collections (PUCs) in MADRaT}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r, echo = FALSE}
+knitr::opts_chunk$set(collapse = TRUE, comment = "#>")
+```
+
+In some use-cases it can be useful to be able to share a unaggrgated version of the data collections by madrat, e.g if a partner want to be able to compute the data collection with a custom aggregation without the need to rerun the whole data processing, or if a snapshot of a data collection should be taken in a form in which it can be later re-used in other aggregations. Portable Unaggregated Collections (PUCs) are the tool to do so.
```
pfuehrlich-pik, line 16:
```suggestion
In some use-cases it can be useful to be able to share an unaggregated version of the data collections created by madrat::retrieveData, e.g. if a partner wants to be able to compute the data collection with a custom aggregation without the need to rerun the whole data processing, or if a snapshot of a data collection should be taken in a form in which it can be later re-used in other aggregations. Portable Unaggregated Collections (PUCs) are the tool to do so. 
```

vignettes/madrat-puc.Rmd
```diff
@@ -0,0 +1,96 @@
...
+
+
+## Basics
+
+The core idea of madrat is to create data processing workflow in a format in which it can be shared and re-used by others. Theoretically that means that a user should just need to have a madrat-package in order to recompute the resulting data collection. In practice this does not always work (e.g. because of problems accessing the source data) or might be impratical for certain applications (e.g. because of high runtimes and/or hardware requirements of the preprocessing). 
+
+In these instances a solution is to share the code (for transparency reason) along with the computed data collection. A drawback of that solution is that the data is already aggregated to a specified regional aggregation, limiting the range of application of such an approach. PUCs represent an intermediate product in which most of the computations have already been performed in advanced but the data still needs to be aggregated to its final aggregation and (potentially) some less time consuming computation steps still have to be done. This can be great for portability of the data collection while maintaining some degree of freedom when it comes to parametrization of the data processing.
```
pfuehrlich-pik, line 23:
```suggestion
In these instances a solution is to share the code (for transparency reasons) along with the computed data collection. A drawback of that solution is that the data is already aggregated to a specified regional aggregation, limiting the range of application of such an approach. PUCs represent an intermediate product in which most of the computations have already been performed in advance but the data still needs to be aggregated to its final aggregation and (potentially) some less time consuming computation steps still have to be done. This can be great for portability of the data collection while maintaining some degree of freedom when it comes to parametrization of the data processing. 
```

vignettes/madrat-puc.Rmd
```diff
@@ -0,0 +1,96 @@
...
+
+In these instances a solution is to share the code (for transparency reason) along with the computed data collection. A drawback of that solution is that the data is already aggregated to a specified regional aggregation, limiting the range of application of such an approach. PUCs represent an intermediate product in which most of the computations have already been performed in advanced but the data still needs to be aggregated to its final aggregation and (potentially) some less time consuming computation steps still have to be done. This can be great for portability of the data collection while maintaining some degree of freedom when it comes to parametrization of the data processing. 
+
+## A typical workflow
+
+By default a PUC is created automatically when a daa processing is launched:
```
pfuehrlich-pik, line 27:
```suggestion
By default a PUC is created automatically when a data processing is launched:
```

vignettes/madrat-puc.Rmd
```diff
@@ -0,0 +1,96 @@
...
+
+
+```{r, echo = TRUE, eval=FALSE}
+library(madrat, quietly = TRUE)
+retrieveData("EXAMPLE", rev = 42, puc = TRUE, extra = "Extra Argument")
+```
+
+In this example the example collection from this package is computed in revision 42. The argument `puc` hereby controls whether the processing should also create a puc-file or not. As `puc = TRUE` is the default setting it does not have to be mentioned here specifically in order to be computed. The `extra` argument is an additional parameter forwarded to the example data collection.
+
+Running this code will create the aggregated collection in the output folder (`getConfig("outputfolder)`) as well as the portable, unaggregated collection in the puc-folder (`getConfig(puc)`).
```
pfuehrlich-pik, line 37:
```suggestion
Running this code will create the aggregated collection in the output folder (`getConfig("outputfolder)`) as well as the portable, unaggregated collection in the puc-folder (`getConfig("pucfolder")`).
```

vignettes/madrat-puc.Rmd
```diff
@@ -0,0 +1,96 @@
...
+
+In this example the puc file has the name `rev42_extra_example_tag.puc` and consists of different components. Every puc-file has the same name structure, which is `rev<revisionNumber>_<selectableArguments>_<collectionName>_<tag>.puc` in which `<revisionNumber>` stands for the revision number, `<selectableArguments>` stands for arguments additional to the regional aggregated which can be selected when aggregating a collection from a puc-file (other arguments cannot be changed as they would require a new puc-file), `<collectionName>` stands for the name of the data collection and `<tag>` stands for a optional name tag, which can be specified in the corresponding `full`-function but which also can be left out.
```
pfuehrlich-pik, line 39:
```suggestion
In this example the puc file name is `rev42_extra_example_tag.puc` and it consists of different components. Every puc-file has the same name structure, which is `rev<revisionNumber>_<selectableArguments>_<collectionName>_<tag>.puc` in which `<revisionNumber>` stands for the revision number, `<selectableArguments>` stands for additional arguments (in addition to the regional aggregation) that can be selected when aggregating a collection from a puc-file (other arguments cannot be changed as they would require a new puc-file), `<collectionName>` stands for the name of the data collection and `<tag>` stands for an optional name tag, which can be specified in the corresponding `full`-function.
```

vignettes/madrat-puc.Rmd
```diff
@@ -0,0 +1,96 @@
...
+
+In our example we have the puc for the data collection "example" in revision 42 and can select the argument "extra" when we aggregate data from the puc file.
```
pfuehrlich-pik, line 41:
```suggestion
In our example we have the puc for the data collection "example" in revision 42 and users can select the argument "extra" when aggregating data from the puc file.
```

vignettes/madrat-puc.Rmd
```diff
@@ -0,0 +1,96 @@
...
+
+In our example we have the puc for the data collection "example" in revision 42 and can select the argument "extra" when we aggregate data from the puc file. 
+
+Aggregating from a puc-file can now happen in two ways:
+
+* on the system where the puc-file has been created it will serve as a snapshot and will be re-used as soon as `retrieveData` is run again with same settings except of the ones which can be changed in the puc-File (e.g. `retrieveData("EXAMPLE", rev=42, puc = TRUE, extra = "Other Argument")`)
+
+* when pushing the puc to someone else one can aggregate the puc-file using the function `pucAggregate` (e.g. `pucAggregate("rev42_extra_example_tag.puc"), extra = "Other Argument")`)
```
pfuehrlich-pik, line 47:
```suggestion
* when giving the puc to someone else they can aggregate the puc-file using the function `pucAggregate` (e.g. `pucAggregate("rev42_extra_example_tag.puc"), extra = "Other Argument")`)
```

vignettes/madrat-puc.Rmd
```diff
@@ -0,0 +1,96 @@
...
+
+In both cases a new aggregated collection will be written into the outputfolder based on the given puc-file.
+
+## Making a madrat preprocessing read for puc-files
```
pfuehrlich-pik, line 51:
```suggestion
## Making a madrat preprocessing ready for puc-files
```

vignettes/madrat-puc.Rmd
```diff
@@ -0,0 +1,96 @@
...
+
+While many parts of the puc-file creation happen automatically, some specific cases require manual tweaking. By default the cache files of the calc-functions called directly by the corresponding full-function are being put into the puc-file. Doing so will later only require to rerun the full-function itself but now underlying calculations as all data of these calculations is already part of the puc-file. However, in some instances
```
pfuehrlich-pik, line 53:
```suggestion
While many parts of the puc-file creation happen automatically, some specific cases require manual tweaking. By default the cache files of the calc-functions called directly by the corresponding full-function are being put into the puc-file. This way only the full-function itself needs to be re-run without running the underlying calculations as all the data of these calculations is already part of the puc-file. However, in some instances
```

vignettes/madrat-puc.Rmd
```diff
@@ -0,0 +1,96 @@
...
+these top-level cache-files are not the ones which should be put into the puc-file (e.g. if these calculations should be recomputed every time a puc-file is being aggregated). Whether a file should be considered for a puc-file or not can be controlled by the return values of a calc-function:
```
pfuehrlich-pik, line 54:
```suggestion
these top-level cache-files are not the ones which should be put into the puc-file (e.g. if these calculations should be recomputed every time a puc-file is being aggregated). Whether a file should be included in a puc-file can be controlled by the return value of a calc-function:
```

vignettes/madrat-puc.Rmd
```diff
@@ -0,0 +1,96 @@
...
+
+```{r, echo = TRUE, eval=FALSE}
+calcExample <- function() {
+return(list(x = data,
+            putInPUC = FALSE))
+}
+```
+
+the calc-function in the shown example uses the return value `putInPUC = FALSE` to overwrite the automatic puc-detection and prevent the cache file from being stored in a puc-file. In a similar fashion `putInPUC = TRUE` would make sure that the file becomes a part of the resulting puc-file.
```
pfuehrlich-pik, line 63:
```suggestion
The calc-function in the shown example uses the return value `putInPUC = FALSE` to overwrite the automatic puc-detection and prevent the cache file from being stored in a puc-file. In a similar fashion `putInPUC = TRUE` would make sure that the file becomes a part of the resulting puc-file.
```

pfuehrlich-pik (approved):
Looks good, I think we can merge after fixing some things in the vignette :slightly_smiling_face:

PR #108, opened 2022-03-07
Enable the use of multiple mappings (usually regions) to which the data should be aggregated

dklein-pik:
If multiple mappings match the data, they will be merged into one mapping before aggregation.


R/calcOutput.R
```diff
@@ -338,19 +338,57 @@ calcOutput <- function(type, aggregate = TRUE, file = NULL, years = NULL, round
...
+    # mappings must have the same length AND the same items as data
+    relFitting <- which(vapply(rel, nrow, FUN.VALUE = integer(1)) == length(items) &
+                        !vapply(columnNameWithItems, identical, character(0), FUN.VALUE = logical(1)))
+
     if (length(relFitting) == 0) stop("Neither getConfig(\"regionmapping\") nor getConfig(\"extramappings\")",
       " contain a mapping compatible to the provided data!")
+
+    # inform about mappings that don't fit and will be omitted
+    omit <- setdiff(names(rel), names(relFitting))
+    if (length(omit) > 0)
```
tscheypidi, line 361: please use curly braces here of write it in  a single line to make clear that only the following line is part of the if-statement

R/calcOutput.R
```diff
@@ -338,19 +338,57 @@ calcOutput <- function(type, aggregate = TRUE, file = NULL, years = NULL, round
...
+      vcat(verbosity = 2, "Ignoring region mapping ", omit, " because it does not fit the data")
+
+    # keep mappings only that fit the data
+    rel <- rel[relFitting]
+
+    # if there is only one fitting mapping make rel a data frame
+    if (length(rel) == 1)
+      rel <- rel[[1]]
```
tscheypidi, line 369: same here, use curly braces to make it more visible what belongs into the if-statement

tscheypidi (changes requested):
looks good, just some minor requests for modifications

tscheypidi (approved):

PR #114, opened 2022-04-07
update calcOutput() add option for significant digits as an alternative to rounding to decimal places

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
The `round` option of `calcOutput()` has proven problematic since data of vastly different magnitude end up in the same `magclass` objects and files.
My case in point is [`f_fedemand.cs4r`](https://github.com/pik-piam/mrremind/blob/8e7973f09ce52a0d704f212b66d539cf76043203/R/fullREMIND.R#L42), which by necessity contains final energy demand of main fuels for large regions and large agglomerations (e.g. coal for all of industry in `CHA` – `fesoi`) as well as marginal fuels for small regions (in terms of economic activity/energy use) and small sub-agglomerations (e.g. electricity for high-temperature heat for chemical production in `SSA` – `feelhth_chemicals`).  These differ by five orders of magnitude (10 EJ vs. 100 TJ), which leaves very little numerical information when rounding to a fixed number of decimal places.
Since some of these number come in prescribed ratios that constitute boundary conditions in REMIND, their limited accuracy lead to a need for rescaling them after data ingestion, causing a difference between input data and model data. (See [here](https://github.com/remindmodel/remind/blob/1e4f3e58724429da9816541a38ebf63bd5d60447/core/datainput.gms#L1479-L1517), recently changed because of different input data handling.)
The solution is straight-forward: add an option to round to a specified number of significant digits instead of decimal places.  This gives much better fidelity for small values in the resulting files.


R/calcOutput.R
```diff
@@ -13,7 +13,10 @@
 #' outputfolder as specified in the config.
 #' @param years A vector of years that should be returned. If set to NULL all
 #' available years are returned.
-#' @param round A rounding factor. If set to NULL no rounding will occur.
+#' @param round Number of decimal places to round to.  Ignored if `NULL`.  See
+#'   [`round()`] for details.
+#' @param signif Number of significant digits to round to.  Ignored if `NULL` or
+#'   if `round` is not `NULL`.  See [`signif()`] for details.
```
tscheypidi, line 19: I would not ignore it if round is being used.

R/calcOutput.R
```diff
@@ -430,8 +434,13 @@ calcOutput <- function(type, aggregate = TRUE, file = NULL, years = NULL, round
   if (!is.null(round)) {
     if (x$class != "magpie") stop("rounding can only be used in combination with x$class=\"magpie\"!")
     x$x <- round(x$x, round)
+  } else if (!is.null(signif)) {
```
tscheypidi, line 437:
```suggestion
  }
  if (!is.null(signif)) {
```

R/calcOutput.R
tscheypidi: but if both can be used at the same time it would be good to indicate here in which order they are being applied

tscheypidi (changes requested):

R/calcOutput.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Seemed like the option to me best preserving the original behaviour.
I don't see any use in using both `round()` and `signif()`, but the result would be the same no matter the order in which they are used.

R/calcOutput.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: I dislike committing from the web interface.  Will force-push from remote.

R/calcOutput.R
tscheypidi: oh true, so ignore my comment about the order. I agree that it is hard to imagine a case where the application of both at the same time would make sense, but I would not preclude the possibility to do so as there are no technical reasons for it.

tscheypidi (approved):

tscheypidi:
@0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q it means that the line coverage slightly decreased. You could add a test for the new argument, but it is also ok for me if you merge now

PR #119, opened 2022-04-28
Generate region hash for extramappings and append to the file name of the input data.

dklein-pik:


R/retrieveData.R
```diff
@@ -73,7 +73,7 @@ retrieveData <- function(model, rev = 0, dev = "", cachetype = "rev", puc = iden
 
     .match <- function(folder, fileType, pattern) {
       match <- dir(path = folder, pattern = paste0(".*\\.", fileType))
-      match <- match[startsWith(match, pattern) & !startsWith(match, paste0(pattern, "_debug"))]
+      match <- match[(startsWith(match, paste0(pattern, "_")) | startsWith(match, paste0(pattern, "."))) & !startsWith(match, paste0(pattern, "_debug"))] # nolint
```
tscheypidi, line 76: why is the `nolint` required here?

R/retrieveData.R
```diff
@@ -100,42 +100,54 @@ retrieveData <- function(model, rev = 0, dev = "", cachetype = "rev", puc = iden
...
 
-  # copy mapping to mapping folder and set config accordingly
-  regionmapping <- toolGetMapping(getConfig("regionmapping"), type = "regional", returnPathOnly = TRUE)
-  mappath <- toolGetMapping(paste0(cfg$regionscode, ".csv"), "regional", error.missing = FALSE, returnPathOnly = TRUE)
-  if (!file.exists(mappath)) {
-    if (!dir.exists(dirname(mappath))) {
-      dir.create(dirname(mappath), recursive = TRUE)
+  # copy mappings to mapping folder and set config accordingly
+  regionmapping <- sapply(c(getConfig("regionmapping"),getConfig("extramappings")), toolGetMapping, type = "regional", returnPathOnly = TRUE) # nolint
+  mappath <- sapply(paste0(cfg$regionscode, ".csv"), toolGetMapping, "regional", error.missing = FALSE, returnPathOnly = TRUE) # nolint
```
tscheypidi, line 112: nolint?

tscheypidi (changes requested):
looks good. Only two things:

* one check in github action currently fails due to a new note (unrelated to your changes). It would be nice if you could fix this as part of your PR (use of `inherits` function for class comparison)
* there are 3 `nolint` statements which I believe you can probably get rid of (e.g. line breaks in case of too long lines of code)

R/retrieveData.R
dklein-pik: Lintr complained about the line lenght. I inserted a line break.

tscheypidi (approved):
looks good

PR #124, opened 2022-07-04
Correct typo, work around inst problems, refactor toolGetMapping

mikapfl:
Hi,

So, originally, I just wanted to correct a typo in the documentation "tmfile" → "tmpfile". But then building madrat was not possible any more due to a newly introduced error in devtools/pkgload. Fixing this led to the linter commanding me to refactor toolGetMapping, which I did.

I *understand* the rules, but they lead to a lot of unexpected work.

Cheers,

Mika


R/toolGetMapping.R
```diff
@@ -24,58 +24,87 @@
...
+  return(paste0(mf, "/", type, "/", name))
+}
+
+.searchNamePackage <- function(name, type, packageName) {
+  tmpfname <- .typedName(name, type)
+  fname <- system.file("extdata", tmpfname, package = packageName)
+  if (fname == "" && !is_dev_package(packageName)) fname <- system.file("inst/extdata", tmpfname, package = packageName)
+  if (fname == "") fname <- system.file("extdata", strsplit(tmpfname, split = "/")[[1]][2], package = packageName)
+  if (fname == "" && !is_dev_package(packageName))
+    fname <- system.file("inst/extdata", strsplit(tmpfname, split = "/")[[1]][2], package = packageName)
```
mikapfl, line 96: Here are the new checks for `is_dev_package` - if it is, `system.file("inst/extdata")` throws an error because stuff under `inst/` is installed at the package top level, so this `system.file` call should never find anything. To be perfectly bug-compatible with the current madrat in production, I introduced the `!is_dev_package` guard. But *most likely* the `system.file("inst/…")` calls were introduced in error and they could be removed instead. That would make things less weird. What do you think?

R/toolGetMapping.R
```diff
@@ -24,58 +24,87 @@
...
   }
-  if (error.missing & !file.exists(as.character(fname))) stop('Mapping "', name, '" not found!')
-  fname <- gsub("/+", "/", fname)
-  if (returnPathOnly) return(fname)
+}
+
+.searchNameLocal <- function(name, type) {
+  if (file.exists(as.character(name))) {
+    return(name)
+  }
```
mikapfl, line 81: This check if a bare `name` exists used to be only present in `where = NULL`, not in `where = "local"`, which also meant that the documentation was wrong (which claimed that `where = NULL` starts by doing the same thing as "local"). I think it is somewhat useful to find a mapping if the full path is given and the type is specified, so I added it to `where = "local"` calls as well.

R/toolGetMapping.R
mikapfl: Interestingly, the `system.file("inst/…")` call(s) were introduced in https://github.com/pik-piam/madrat/commit/6f9d872425bc6aacfaba1a2c2947d82a033b5f60 due to a "bug in `system.file`", so apparently 5 years ago it was necessary to ask for "inst/…"? But when was `system.file` fixed? Can we drop this workaround or do we need to pile the other workaround onto this one?

tscheypidi (approved):

PR #127, opened 2022-07-18
support for bilateral data + fixes

tscheypidi:
* added bilateral data support in `calcOutput`
* added a check in `addMapping`
* added argument `strict` in `pucAggregate` and set it to `FALSE` by default
* performance tweak in `toolAggregate` for instances in which all values in a column are NA

### DIFF
--- a/R/calcOutput.R
+++ b/R/calcOutput.R
@@ -119,6 +119,17 @@ calcOutput <- function(type, aggregate = TRUE, file = NULL, years = NULL, # noli
 
   # read region mappings check settings for aggregate
   if (aggregate != FALSE) {
+    .bilateralMapping <- function(mapping) {
+      out <- NULL
+      for (i in seq_len(ncol(mapping))) {
+        tmp <- expand.grid(mapping[[i]], mapping[[i]])
+        out <- cbind(out, do.call(paste, args = c(tmp, list(sep = "."))))
+      }
+      out <- as.data.frame(out, stringsAsFactors = FALSE)
+      colnames(out) <- colnames(mapping)
+      return(out)
+    }
+
     rel <- list()
     relNames <- NULL
     for (r in c(getConfig("regionmapping"), getConfig("extramappings"))) {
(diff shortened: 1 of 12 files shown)


R/calcOutput.R
```diff
@@ -119,6 +119,17 @@ calcOutput <- function(type, aggregate = TRUE, file = NULL, years = NULL, # noli
 
   # read region mappings check settings for aggregate
   if (aggregate != FALSE) {
+    .bilateralMapping <- function(mapping) {
+      out <- NULL
+      for (i in seq_len(ncol(mapping))) {
```
pfuehrlich-pik, line 124: Can we iterate `mapping` directly instead of this indexed based approach?

R/calcOutput.R
```diff
@@ -129,6 +140,9 @@ calcOutput <- function(type, aggregate = TRUE, file = NULL, years = NULL, # noli
       if (is.null(rel[[r]]$global)) {
         rel[[r]]$global <- "GLO"  # add global column
       }
+      if (nrow(rel[[r]]) < 250) {
```
pfuehrlich-pik, line 143: Why 250? Maybe add a comment to clarify?

tests/testthat/test-calcOutput.R
```diff
@@ -269,6 +283,27 @@ test_that("Aggregation works", {
   })
 })
 
+test_that("Bilateral aggregation works", {
+  calcBilateral <- function() {
+    map <- toolGetMapping("regionmappingH12.csv")
+    tmp <- expand.grid(map[[2]], map[[2]], stringsAsFactors = FALSE)
```
pfuehrlich-pik, line 289: Is there a reason why stringsAsFactors is set explicitly here but not in calcOutput (line 125)?
```suggestion
    expandedMapping <- expand.grid(map[[2]], map[[2]], stringsAsFactors = FALSE)
```

tscheypidi: could catch...I will add it there as well

PR #129, opened 2022-08-05
fix vcat 'list cannot be handled by cat'

pfuehrlich-pik:
Calling `vcat(1, list())` was crashing, this PR fixes that. The key was just to deparse lists into character (see first change in vcat.R), the rest is just reducing cyclomatic complexity and linter stuff.


mikapfl (approved):
lgtm, thanks!

PR #135, opened 2022-09-29
Add two alternative country names for Macedonia.

Loisel:
Only relevant for the table that supplies `toolCountry2isocode`.

### DIFF
--- a/inst/extdata/country2iso.csv
+++ b/inst/extdata/country2iso.csv
@@ -124,6 +124,8 @@ Eritrea;ERI
 Estonia;EST
 Ethiopia;ETH
 FYR of Macedonia;MKD
+FYROM;MKD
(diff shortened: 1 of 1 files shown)


giannou:
The country is now called North Macedonia, or Republic of North Macedonia

Loisel:
> The country is now called North Macedonia, or Republic of North Macedonia

Ok thanks. Are you saying this because of the PR title or do you think something should be changed in the PR itself?

giannou:
> > The country is now called North Macedonia, or Republic of North Macedonia
> 
> Ok thanks. Are you saying this because of the PR title or do you think something should be changed in the PR itself?

The PR is fine (it maps _alternative_ names to the official country code, so nothing wrong there), I was referring to the title :)

giannou (approved):

Issue #136, opened 2022-09-30
toolCountryFill should preserve comment


Loisel:
`toolCountryFill` does not preserve the comment of a magpie object. If this is not intended, the documentation should be explicit.

To reproduce:
```
t <- new.magpie(years=2020, fill=1)
getComment(t) <- "test comment"
x <- toolCountryFill(t)
getComment(x)
```

Loisel:
solved in #137

PR #137, opened 2022-09-30
Preserve Comment in toolCountryFill

Loisel:
Preserve the comment of a magpie object when returned from `toolCountryFill`. Addresses issue #136


tscheypidi (approved):
looks good, thanks for fixing it!

PR #141, opened 2022-10-28
fixed vcat behavior

tscheypidi:
- vcat should now properly react on suppressMessages or suppressWarnings statements
- removed diagnostics_full.log as it was containing partly misleading (supposedly) suppressed notes/warnings, use a higher verbosity instead to compute the corresponding, detailed information
- fixed linter warnings


pfuehrlich-pik (approved):
Cool, looks good!

PR #146, opened 2022-12-20
support returning non-magpie from readSource

pfuehrlich-pik:


R/readSource.R
```diff
@@ -4,41 +4,41 @@
...
+#' for huge data sets where reading in the whole data set might be impractical or even infeasible.
 #' @param convert Boolean indicating whether input data conversion to
 #' ISO countries should be done or not. In addition it can be set to "onlycorrect"
 #' for sources with a separate correctXXX-function.
-#' @return magpie object with the temporal and data dimensionality of the
-#' source data. Spatial will either agree with the source data or will be on
-#' ISO code country level depending on your choice for the argument "convert".
-#' @author Jan Philipp Dietrich, Anastasis Giannousakis, Lavinia Baumstark
+#' @return The read-in data, usually a magpie object. Technically speaking
+#' `if (is.list(result)) result$x else result` will be returned, where `result` is the
```
tscheypidi, line 20: not sure if that part is easy to understand for everyone

R/readSource.R
```diff
@@ -4,41 +4,41 @@
...
+#' return value of the wrapped function. The temporal and data dimensionality
+#' should match the source data. The spatial dimension should either match the source data or,
+#' if the convert argument is set to TRUE, should be on ISO code country level. For magpie objects
+#' magclass::clean_magpie is run and if convert = TRUE ISO code country level is checked.
```
tscheypidi, line 24: why do you mention `clean_magpie` here? If you want to add more details I would add them as a note, not as part of the description of the return value

R/readSource.R
```diff
@@ -96,37 +96,62 @@ readSource <- function(type, subtype = NULL, subset = NULL, convert = TRUE) { #
...
       setWrapperInactive("wrapperChecks")
     })
 
-    if (!is.magpie(x)) {
-      stop('Output of function "', functionname, '" is not a MAgPIE object!')
-    }
-    if (prefix == "convert") {
-      .testISO(getItems(x, dim = 1.1), functionname = functionname)
+    # ensure we are always working with a list with entries "x" and "class"
+    xList <- if (is.list(x)) x else list(x = x)
```
tscheypidi, line 128:
```suggestion
    xList <- if (is.list(x)) x else list(x = x, class = "magpie")
```

R/readSource.R
```diff
@@ -96,37 +96,62 @@ readSource <- function(type, subtype = NULL, subset = NULL, convert = TRUE) { #
...
+
+    # return list is only available for read, not for correct/convert
+    if (prefix == "read") {
+      if (!is.list(x)) {
+        xList$class <- "magpie"
+      }
```
tscheypidi, line 134:
```suggestion
```

R/readSource.R
```diff
@@ -96,37 +96,62 @@ readSource <- function(type, subtype = NULL, subset = NULL, convert = TRUE) { #
...
 
-    if (!is.magpie(x)) {
-      stop('Output of function "', functionname, '" is not a MAgPIE object!')
-    }
-    if (prefix == "convert") {
-      .testISO(getItems(x, dim = 1.1), functionname = functionname)
+    # ensure we are always working with a list with entries "x" and "class"
+    xList <- if (is.list(x)) x else list(x = x)
+
+    # return list is only available for read, not for correct/convert
```
tscheypidi, line 130: why is the return list only available for read? it should be also available in the other two cases I think

R/readSource.R
```diff
@@ -96,37 +96,62 @@ readSource <- function(type, subtype = NULL, subset = NULL, convert = TRUE) { #
...
+      if (magclass::is.magpie(xList$x)) {
+        .testISO(magclass::getItems(xList$x, dim = 1.1), functionname = functionname)
+      } else {
+        vcat(2, "Non-magpie objects are not checked for ISO country level.")
+      }
     }
-    cachePut(x, prefix = prefix, type = type, args = args)
-    return(x)
+
+    cachePut(xList$x, prefix = prefix, type = type, args = args)
```
tscheypidi, line 153:
```suggestion
    cachePut(xList, prefix = prefix, type = type, args = args)
```

tests/testthat/test-caching.R
```diff
@@ -1,11 +1,7 @@
-globalassign <- function(...) {
-  for (x in c(...)) assign(x, eval.parent(parse(text = x)), .GlobalEnv)
-}
-
 test_that("Caching works", {
   calcCacheExample <- function() return(list(x = as.magpie(1), description = "-", unit = "-"))
   globalassign("calcCacheExample")
-  localConfig(globalenv = TRUE, ignorecache = FALSE, .verbose = FALSE)
```
tscheypidi, line 8: why is `globalenv = TRUE ` not required anymore? isn`t `calcCacheExample` still assigned to it?

tscheypidi (changes requested):

R/readSource.R
pfuehrlich-pik: Yeah I know, I was really struggling to find a good explanation. Do you have a suggestion?

R/readSource.R
pfuehrlich-pik: You mean as `vcat(2, ...)`? `clean_magpie` changes the return value, so I think it makes sense here, no?

tscheypidi: no, I meant in the documentation. There is a category `@details` which could be used for it (I thought it was `@note` which is why my wording was misleading, sorry for that).

tests/testthat/test-caching.R
pfuehrlich-pik: setup.R sets `globalenv = TRUE` for all tests.

R/readSource.R
tscheypidi: I would just skip the sentence starting with `technically speaking`

pfuehrlich-pik: I think it's important to tell the user that if they return a list from their read function they will not get that list when calling `readSource`, unless they pass `supplementary = TRUE`

R/readSource.R
pfuehrlich-pik: ok sure

R/readSource.R
pfuehrlich-pik: I think now that `supplementary` is available and documented it's clear without the sentence, so I removed it as you suggested.

tscheypidi (approved):
excellent work!

tscheypidi (approved):
update looks good

PR #149, opened 2023-01-09
Compatibility to old cache files

pfuehrlich-pik:
The latest madrat update introduced a new cache file format (list) which could lead to problems when loading old cache files with the new madrat via forcecache. This PR solves these issues.


tscheypidi (approved):
looks good

PR #166, opened 2023-04-25
add coding rule to madrat tutorial

fbenke-pik:


LaviniaBaumstark:
But how would you ensure reproducability of old data? We should be able to reproduce input data based on R-packageversions, means that a read* function needs to be adjusted to read in correlated input data -> versioning of sources. Could also be solved by subfolders of the source folder.

tscheypidi:
yes, but that should be managed via different subtypes in download-functions, not by manually adjusting the source folder structure. Normally, every source function should come with a download function and the user should not manually adjust the contents of the source folders

PR #169, opened 2023-05-08
let PUC creation use madrat specific tmpdir

tscheypidi:
- make use of getConfig("tmpfolder") instead of tempdir() for temporary files related to puc files


FelicitasBeier (approved):
Looks good to me.

PR #171, opened 2023-06-29
replaced shortened inv = TRUE arguments with invert = TRUE (bugfix)

tscheypidi:


CRAN-RELEASE
```diff
@@ -0,0 +1,2 @@
+This package was submitted to CRAN on 2019-12-17.
+Once it is accepted, delete this file and tag the release (commit ebf7bcc9db).
```
pfuehrlich-pik, line 2: Why do we need this file now?

R/toolISOhistorical.R
```diff
@@ -367,17 +367,18 @@ toolISOhistorical <- function(m, mapping = NULL, additional_mapping = NULL, over
            m[getItems(mTr, dim = 1), getYears(mTr), ] <- ifelse(is.na(m[getItems(mTr, dim = 1), getYears(mTr), ]),
                                                                      mTr, m[getItems(mTr, dim = 1), getYears(mTr)])
          if (length(addM) > 0) {
-          m[is.na(m)] <- 0 }
+          m[is.na(m)] <- 0
+ }
```
pfuehrlich-pik, line 371: broken indentation

CRAN-RELEASE
tscheypidi: we don't. I added it by accident. I removed it now again

R/toolISOhistorical.R
tscheypidi: fixed

pfuehrlich-pik (approved):

PR #176, opened 2023-07-21
added filter argument in visualizeDependencies

tscheypidi:


R/visualizeDependencies.R
```diff
@@ -2,25 +2,26 @@
 #'
 #' Creates a graphical visualization of dependencies between functions in the mr-universe.
 #'
-#' @param functions function or vector of functions to be analyzed
+#' @param ... function or vector of functions to be analyzed
```
deleip, line 5: this can also be multiple arguments now, right? Then it should be "functions or vector of functions to be analyzed" I would say. Or "function(s)"

deleip (approved):
Cool functionality, thanks for adding! Just one small comment in the parameter description

R/visualizeDependencies.R
tscheypidi, line 5:
```suggestion
#' @param ... function(s) to be analyzed
```

PR #180, opened 2023-08-18
Add warning for unset 'where' argument in toolGetMapping function

whitehacker:
As it was discussed [here](https://github.com/pik-piam/discussions/discussions/51#discussioncomment-6037935), the changes are made to display a warning to alert users when the `where` argument is not set in `toolGetMapping()` function.


tests/testthat/test-toolGetMapping.R
```diff
@@ -38,4 +38,12 @@ test_that("toolGetMapping works", {
...
   expect_error(toolGetMapping("test.rda", where = "mappingfolder"), "did not contain a object")
   expect_error(toolGetMapping("test.xyz"), "Unsupported filetype")
+
+  readTest <- function() {
+    toolGetMapping("test.csv", type = "regional")
+    return(as.magpie(1))
+  }
+  globalassign("readTest")
+  expect_warning(readSource("Test", convert = "onlycorrect"),
+                "No correct function for Test could be found. Set convert to FALSE.")
```
pfuehrlich-pik, line 48: This test is not testing the warning you added. I think passing `convert = FALSE` or not passing `convert` at all might already be enough to trigger your warning.

tests/testthat/test-toolGetMapping.R
```diff
@@ -38,4 +38,12 @@ test_that("toolGetMapping works", {
   writeLines("abc", paste0(getConfig("mappingfolder"), "/test.xyz"))
   expect_error(toolGetMapping("test.rda", where = "mappingfolder"), "did not contain a object")
   expect_error(toolGetMapping("test.xyz"), "Unsupported filetype")
+
+  readTest <- function() {
+    toolGetMapping("test.csv", type = "regional")
```
pfuehrlich-pik, line 43: If you use one of the already existing mappings you can omit the somewhat large new test.csv file.
```suggestion
    toolGetMapping("regionmappingH12.csv", type = "regional")
```

pfuehrlich-pik (changes requested):

tests/testthat/test-toolGetMapping.R
whitehacker: I tried both options, but the tests failed.

fbenke-pik: In addition to Pascal's comment, this test is not checking the right warning.

R/toolGetMapping.R
```diff
@@ -31,6 +31,16 @@ toolGetMapping <- function(name, type = NULL, where = NULL,
                            returnPathOnly = FALSE, activecalc = NULL) {
 
   setWrapperInactive("wrapperChecks")
+  if (isWrapperActive("wrapperChecks")) {
```
fbenke-pik, line 34: Why are you testing this condition? As this wrapper is being activated in the line before, your new code portion is never executed.

inst/extdata/test.csv
```diff
@@ -0,0 +1,981 @@
+;x
```
fbenke-pik, line 1: You do not need such a file for your test. Please remove.

R/toolGetMapping.R
fbenke-pik: Once you remove this line, a lot of tests fail. 

As a firsts step, you have to deal with the instances where the tests call madrat functions like `madrat::regionscode` which currently call `toolGetMapping` without a where argument and therefore now produce the new warning. Here, I am not sure what the right approach is: either add the `where` argument or enhance the check to not throw the warning when called from within madrat functions. @pfuehrlich-pik what do you think?

There might be other reasons why the test fail, but we can look into them once this reason has been removed.

pfuehrlich-pik:
> Btw, you have two WIP commits. Please merge the three commits into one for a clean git history once everything else has been solved. (I assume, Pascal and Jan would appreciate it).

I'm fine with WIP commits, I would rather not mess with the history. Clean git history is not really a priority for madrat I believe.

R/toolGetMapping.R
pfuehrlich-pik: Passing this question on to @tscheypidi

tscheypidi: the `isWrapperActive` should exactly achieve that the checks are only active outside of madrat functions. What happens if the checks are being performed before the `wrapperChecks`-wrapper is being deactived?

fbenke-pik (approved):

tscheypidi (approved):

PR #185, opened 2023-10-13
fix codelabels in pucAggregate

pfuehrlich-pik:
relevant changes are in https://github.com/pik-piam/madrat/commit/bb3871c038b3e5d709269f7f705040abde51d51c
I think the reason why codelabels were sometimes not working is because the tgz was generated from an existing puc. The puc includes an renv which is used to re-run the full function in a new R session, and codelabels were not moved over to this new R session. This PR fixes that.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '7208614'
+ValidationKey: '7228624'
 AcceptedWarnings:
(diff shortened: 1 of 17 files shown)


deleip (approved):
Looks good to me. Thanks for fixing this :)

FelicitasBeier (approved):
Cool, thanks Pascal!

PR #188, opened 2023-10-25
Extra attempt to find mapping with toolGetMapping

robinhasse:
Okay this is a tricky one and there is probably a more elegant solution for this, but I didn't dig that deep. So I start with a bug report: 

### Description

`toolGetMapping` is designed to also be able to load mapping from packages that are installed via `devtools::load_all` but this sometimes fails.

#### mr-package under development & madrat installed
- I add a new mapping to `mredgebuildings`
- I load build `mredgebuildings` and load it via `library` 

--> mapping found

#### mr-package under development & madrat installed
- I add a new mapping to `mredgebuildings`
- I load `mredgebuildings` via `devtools::load_all` 

--> mapping not found

#### mr-package under development & madrat under development
The problem is not reproducible if madrat is loaded via `load_all`
- I add a new mapping to `mredgebuildings`
- I load `mredgebuildings` via `devtools::load_all`
- I load `madrat` via `devtools::load_all` (unchanged code though)

--> mapping found

So when fixing this it seems important to me to always build `madrat` when testing the fix.

### System

- Windows
- R 4.3.1

## Bug fix

The function performs several attempts to find the file if the ones before were unsuccessful.

https://github.com/pik-piam/madrat/blob/4f0688ddf39907d03d57561fa5eba3d5027d061f/R/toolGetMapping.R#L97-L105

- My problem described above is fixed if line 100 was executed, which does not happen as `is_dev_package(pkgName)` is `TRUE`
- So if none of the attempts until here worked I repeat line 100 also for development packages
- However, there are cases when this would throw an error (something along the lines "file path cannot start with inst") but I don't fully know when this happens. So I put the additional attempt inside a `tryCatch`.
- This works fine as it only increases the chance to find the mapping and defaults to what the function did before if it cannot. If you know why that is, we could be more specific in the if-condition and avoid the `tryCatch`.
- Additionally, I changed the lines that search without the `type` to use `name` directly. If `type` was null, these line couldn't find anything as file path would contain NA which always results in returning `""`

### DIFF
--- a/R/toolGetMapping.R
+++ b/R/toolGetMapping.R
@@ -65,7 +65,7 @@ toolGetMapping <- function(name, type = NULL, where = NULL,
       packages <- c(fp, grep(fp, packages, invert = TRUE, value = TRUE))
     }
     for (i in packages) {
-      out <- .searchNamePackage(name, type = type, packageName = i)
+      out <- .searchNamePackage(name, type = type, pkgName = i)
       if (out != "") {
         fname <- out
       }
@@ -78,7 +78,7 @@ toolGetMapping <- function(name, type = NULL, where = NULL,
   } else if (where == "local") {
     return(.searchNameLocal(name = name, type = type))
   } else {
-    return(.searchNamePackage(name = name, type = type, packageName = where))
+    return(.searchNamePackage(name = name, type = type, pkgName = where))
   }
 }
 
@@ -94,13 +94,16 @@ toolGetMapping <- function(name, type = NULL, where = NULL,
   return(paste0(mf, "/", type, "/", name))
 }
 
-.searchNamePackage <- function(name, type, packageName) {
+.searchNamePackage <- function(name, type, pkgName) {
   tmpfname <- .typedName(name, type)
-  fname <- system.file("extdata", tmpfname, package = packageName)
-  if (fname == "" && !is_dev_package(packageName)) fname <- system.file("inst/extdata", tmpfname, package = packageName)
-  if (fname == "") fname <- system.file("extdata", strsplit(tmpfname, split = "/")[[1]][2], package = packageName)
-  if (fname == "" && !is_dev_package(packageName))
-    fname <- system.file("inst/extdata", strsplit(tmpfname, split = "/")[[1]][2], package = packageName)
+  fname <- system.file("extdata", tmpfname, package = pkgName)
+  if (fname == "" && !is_dev_package(pkgName)) fname <- system.file("inst", "extdata", tmpfname, package = pkgName)
+  if (fname == "")                             fname <- system.file("extdata",         name,     package = pkgName)
+  if (fname == "" && !is_dev_package(pkgName)) fname <- system.file("inst", "extdata", name,     package = pkgName)
+  if (fname == "") {
+    fname <- tryCatch(system.file("inst", "extdata", tmpfname, package = pkgName),
+                      error = function(e) "")
+  }
   return(fname)
 }
 
@@ -108,21 +111,18 @@ toolGetMapping <- function(name, type = NULL, where = NULL,
   if (is.null(type)) {
     return(name)
   }
-  return(paste0(type, "/", name))
+  return(file.path(type, name))
 }
 
 .readMapping <- function(fname) {
   filetype <- tolower(file_ext(fname))
   if (filetype == "csv") {
-    if (grepl(pattern = ";", x = readLines(fname, 1))) {
-      return(read.csv(fname, sep = ";", stringsAsFactors = FALSE, comment.char = "*"))
-    } else {
-      return(read.csv(fname, sep = ",", stringsAsFactors = FALSE, comment.char = "*"))
-    }
+    sep <- if (grepl(pattern = ";", x = readLines(fname, 1))) ";" else ","
+    return(read.csv(fname, sep = sep, stringsAsFactors = FALSE, comment.char = "*"))
   } else if (filetype == "rda") {
     data <- NULL
     load(fname)
-    if (is.null(data)) stop(fname, " did not contain a object named \"data\"!")
+    if (is.null(data)) stop(fname, " did not contain an object named \"data\"!")
     return(data)
   } else if (filetype == "rds") {
     return(readRDS(fname))
--- a/R/toolGetMapping.R
+++ b/R/toolGetMapping.R
@@ -65,7 +65,7 @@ toolGetMapping <- function(name, type = NULL, where = NULL,
       packages <- c(fp, grep(fp, packages, invert = TRUE, value = TRUE))
     }
     for (i in packages) {
-      out <- .searchNamePackage(name, type = type, pkgName = i)
+      out <- .searchNamePackage(name, type = type, packageName = i)
       if (out != "") {
         fname <- out
       }
@@ -78,7 +78,7 @@ toolGetMapping <- function(name, type = NULL, where = NULL,
   } else if (where == "local") {
     return(.searchNameLocal(name = name, type = type))
   } else {
-    return(.searchNamePackage(name = name, type = type, pkgName = where))
+    return(.searchNamePackage(name = name, type = type, packageName = where))
   }
 }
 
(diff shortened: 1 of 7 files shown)


pfuehrlich-pik:
Did you do any kind of extensive test, e.g. run a preprocessing with this @robinhasse ?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
I will report what I found when I tried to help Robin circumnavigate his problem, and suggest a solution that might be less confusing in the future.

> #### mr-package under development & madrat installed
> 
> * I add a new mapping to `mredgebuildings`
> * I load build `mredgebuildings` and load it via `library`
> 
> → mapping found

Because it is in the regular package library, nothing extraordinary.

Changing the order here …

> #### mr-package under development & madrat under development
> 
> The problem is not reproducible if madrat is loaded via `load_all`
> 
> * I add a new mapping to `mredgebuildings`
> * I load `mredgebuildings` via `devtools::load_all`
> * I load `madrat` via `devtools::load_all` (unchanged code though)
> 
> → mapping found

The `devtools` package installs [a shim](https://github.com/r-lib/pkgload/blob/681b36370887b5925304373edbb5e9ea9a87a6dd/R/shims.R#L52-L117) for `base::system.file()`, that is checking the `inst/` subdirectory of the directory the package was loaded from (as well as the directory itself).

> #### mr-package under development & madrat installed
> 
> * I add a new mapping to `mredgebuildings`
> 
> * I load `mredgebuildings` via `devtools::load_all`
> 
> → mapping not found

Because `madrat` was loaded normally, it is binding to `base::system.file()` and not to `pkgload:::shim_system.file()`, and that defaults (via `find.package()`) to return the base directory from where the package was loaded, not the `inst/` subdirectory, when the package is not found in any library (see [this comment](https://mattermost.pik-potsdam.de/rd3/pl/xhpj44r47tb4dg5otssxjuhwoc)).


```
> library(madrat)
Loading required package: magclass

> devtools::load_all()
ℹ Loading mrremind

> system.file(package = 'mrremind')
[1] "/home/pehl/PIK/pik-piam/mrremind/inst"

> base::system.file(package = 'mrremind')
[1] "/home/pehl/PIK/pik-piam/mrremind"

> identical(system.file, pkgload:::shim_system.file)
[1] TRUE
```

So there are four cases to distinguish, and only one needs special attention:
- `madrat` in library, package in library → `base::system.file()` works
- `madrat` via `load_all()`, package via `load_all()` → `pkgload:::shim_system.file()` works
- `madrat` via `load_all()`, package in libarary → package not in `dev_packages()` [[↑]](https://github.com/r-lib/pkgload/blob/681b36370887b5925304373edbb5e9ea9a87a6dd/R/shims.R#L60), `pkg_load:::shim_system.file()` works
- `madrat` in library, package via `load_all()`
  - can be detected via `is_dev_package()`
  - if so, `base::system.file(package = <package>)` will return the base directory where the package was loaded from via `load_all()`
  - `fname` is then just `file.path(base::system.file(package = packageName), 'inst', tmpfname)`

```
.searchNamePackage <- function(name, type, packageName) {
  tmpfname <- .typedName(name, type)
  
  if (is_dev_package(packageName)) {
    fname <- file.path(base::system.file(package = packageName), "inst", 
                       tmpfname)
  } else {
    fname <- system.file("extdata", tmpfname, package = packageName)
  }
}
```

I am honestly at a loss what all the special cases in `.searchNamePackage` are for or under what circumstances they are supposed to kick in.  If they are needed, this solution obviously will not work, but the code would benefit from commenting.

robinhasse:
Thanks @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q for also digging into this. But there is also the case of `packageName` is `"madrat"` and `is_dev_package(packageName)` is `TRUE`. For example when building/testing madrat. I haven't tested your code above but this was a special case for me that could cause problems. Your code looks much cleaner but I think that the lines 101 and 103 in the original code are there to look in the parent folder (i.e. `extdata`) of the type folder. This is missing in you code.

robinhasse:
> Did you do any kind of extensive test, e.g. run a preprocessing with this @robinhasse ?

No, but due to the if clause in each line, I'm pretty confident that the new code should never fail when the old one worked. Testing it to be sure would still be valuable, you're right.

robinhasse:
> I think that the lines 101 and 103 in the original code are there to look in the parent folder (i.e. `extdata`) of the type folder.

On the other hand I don't think they could have worked with `type = NULL`, so this feature only worked if you didn't expect it to.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> But there is also the case of `packageName` is `"madrat"` and `is_dev_package(packageName)` is `TRUE`. For example when building/testing madrat. I haven't tested your code above but this was a special case for me that could cause problems.

```
> devtools::load_all('~/PIK/pik-piam/madrat/')
ℹ Loading madrat
Loading required package: magclass
> pkgload:::shim_system.file(package = 'madrat')
[1] "/home/pehl/PIK/pik-piam/madrat/inst"
> file.path(base::system.file(package = 'madrat'), 'inst') 
[1] "/home/pehl/PIK/pik-piam/madrat/inst"
```
Just works&trade;.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Let's see.  For line 101 (and 103) to kick in, lines 99 and 100 must turn up empty.

- Line 99 returns empty -> `tmpfname` does not exist under `"extdata"`.
- For line 100 to return something, `packageName` most not have been loaded via `load_all()` (for `is_dev_package(packageName)` to return `FALSE`) _and_ `packageName` must not have been loaded from the library.  No idea how that is going to work.  So lets assume it's empty, too.
- Then on line 101, `system.file("extdata", strsplit(tmpfname, split = "/")[[1]][2], package = packageName)` looks for something named like the second thing in the `tmpfname` path.  Not really save, but I think the intention is to look outside of the `type` directory if the file was not found inside the (possibly non-existent) `type` directory.
  In my opinion, that should just fail.  People ought to know where their files are.  Since mappings where moved to packages rather recently, there should also not be some huge mess of legacy code depending on this.
- Same for line 103, except that I do not see how it ever could execute, just like line 100.

tscheypidi:
please ping me if you have come to a conclusion how to adjust it

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> please ping me when you have come to a conclusion how to adjust it

I think the question you, @tscheypidi, need to answer before this can be concluded is this:
    
"Is picking the second element from the combined type-name path and looking for it in the `extdata/` directory of the package if the type-name path was not found an intended part of the `madrat` that should be kept or not?"

tscheypidi:
Yes, this is intended. I don't recall anymore under which circumstances this becomes relevant, but I know that there are cases for which this is needed.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
@robinhasse :point_right: https://github.com/robinhasse/madrat/pull/1
Works fine locally.  For good measure there's a REMIND input data job with that `madrat` version (`/p/tmp/pehl/preprocessing-remind/log-2023-11-03-28364115.out`).

robinhasse:
Thanks Michaja, now it is looking good and it is well documented. Feel free to merge it. It's essentially your PR ;)

pfuehrlich-pik:
please switch to a non-development version number (not ending in .9001)

R/toolGetMapping.R
```diff
@@ -95,34 +95,44 @@ toolGetMapping <- function(name, type = NULL, where = NULL,
...
-  tmpfname <- .typedName(name, type)
-  fname <- system.file("extdata", tmpfname, package = packageName)
-  if (fname == "" && !is_dev_package(packageName)) fname <- system.file("inst/extdata", tmpfname, package = packageName)
-  if (fname == "") fname <- system.file("extdata", strsplit(tmpfname, split = "/")[[1]][2], package = packageName)
-  if (fname == "" && !is_dev_package(packageName))
-    fname <- system.file("inst/extdata", strsplit(tmpfname, split = "/")[[1]][2], package = packageName)
-  return(fname)
+  packageLocation <- if (is_dev_package(packageName)) {
+    # If <packageName> was attached using devtool::load_all(), system.file()
+    # might nor might not be shimmed with pkgload:::shim_system.file()
```
pfuehrlich-pik, line 100:
```suggestion
    # If <packageName> was attached using devtools::load_all(), system.file()
    # might or might not be shimmed with pkgload:::shim_system.file()
```

pfuehrlich-pik (approved):

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> please switch to a non-development version number (not ending in .9001)

Sure.  Once the test is done.

PR #190, opened 2023-11-08
add writeArgs to calcOutput

fbenke-pik:


tscheypidi:
`writeArgs` should not be limited to write.report, this should work for all cases in which the file argument is set to something. Aside from that it looks good.

fbenke-pik:
Not sure how I could test when `x` is not a magclass object (causes `saveRDS` to be called)

tscheypidi (approved):
looks good

PR #191, opened 2023-11-10
handle writing all-NA magclass objects in calcOutput

fbenke-pik:
For some aggregations and calc functions, we produce all-NA magclass objects (e.g. H12 for a dataset that only has values for DEU). In this case, write.report crashes during the conversion to a data table. 

In this PR, calcOutput checks for all-NA magclass objects after regional aggregation and before trying to write to a mif and throws a warning for all-NAs.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '7275680'
+ValidationKey: '7297199'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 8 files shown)


fbenke-pik:
Alternatively, I could look into handling this directly in write.report, if that is the preferred approach.

tscheypidi:
yes, please do it in `write.report` itself as this seems more to be a bug in that function than an issue in madrat

PR #206, opened 2024-05-17
Add new tool function

johanneskoch94:
The idea here is to create a central parameter/function with which we control/set the monetary unit, throughout the magpie/remind pre-processing. 
In effect, the crucial part is the monetary unit's **base year** (our choice of currency, the $, in PPP or MER, not really being up for debate).

Ideally, this function will be called every time monetary values are converted, providing the unit in which the values are to be converted to. The conversion process itself should rely on `GDPuc::convertGDP` (since this function/package will be used for all monetary conversions, the naming should perhaps be revised...). To make the whole process as easy as possible, the proposed function, here called `toolGetUnitDollar`, would ideally return the exact string that can be passed to `GDPuc::convertGDP`'s `unit_out` argument.

Questions:

- Is this package the right place to introduce a function like this?
- Should we take this opportunity and create the framework for a more general `toolGetUnit` function?
- Where should the units be defined? In this case, the base year is defined within the tool function itself. Perhaps a special, external file should be used instead? Where all units are defined?

### DIFF
--- a/R/toolGetUnitDollar.R
+++ b/R/toolGetUnitDollar.R
@@ -1,0 +1,8 @@
+toolGetUnitDollar <- function(returnOnlyBase = FALSE, inPPPP = TRUE) {
+  base <- 2017
+  if (returnOnlyBase) {
+    return(base)
+  }
+  pppOrMer <- if (inPPPP) " Int$PPP" else " US$MER"
+  paste0("constant ", base, pppOrMer)
+}


johanneskoch94:
Another point: we would of course have to agree on a unique formulation for the unit of monetary values. No more 2005$, 05$, 2005 US$, US$05, US$ 2005, constant 2005US$, USD05, 2005 Int$, and so on.... Perhaps a choice was already made in `piamInterfaces` that we could take over? Otherwise, the formulation accepted by `GDPuc`, i.e. "constant 2017 Int$PPP" or "constant 2017 US$MER", could be used. It wouldn't be the most concise way of putting it, but the most explicit, avoiding any possible confusion...

tscheypidi:
this is too application specific to be put in madrat. I guess mrdrivers could be a better spot

LaviniaBaumstark:
If this function only defines teh dollar unit, I agree with JPD and would put it in mrdrivers. Let's stat with this. Maybe later we will decide to implement a more general default-unit-function for madrat and can use this dollar-function as a starting point.

LaviniaBaumstark:
> Another point: we would of course have to agree on a unique formulation for the unit of monetary values. No more 2005$, 05$, 2005 US$, US$05, US$ 2005, constant 2005US$, USD05, 2005 Int$, and so on.... Perhaps a choice was already made in `piamInterfaces` that we could take over? Otherwise, the formulation accepted by `GDPuc`, i.e. "constant 2017 Int$PPP" or "constant 2017 US$MER", could be used. It wouldn't be the most concise way of putting it, but the most explicit, avoiding any possible confusion...

@orichters @fbenke-pik do you know a good standard from `piamInterfaces`?

orichters:
REMIND uses `US$2010` quite consistently, whereas MAgPIE had some more variety, but already consolidated a bit since I told them about it. But piamInterfaces basically treats [all those versions](https://github.com/pik-piam/piamInterfaces/blob/master/R/areUnitsIdentical.R#L35-L36) as mutually identical, and we could easily add a new version if needed. The new community standard [will probably be](https://github.com/IAMconsortium/common-definitions/blob/main/definitions/variable/macro-economy/gdp.yaml) `billion USD_2010/yr` to avoid issues with the special character `$`. Not sure whether PPP and MER should be part of the unit, but maybe it is helpful.

fbenke-pik:
Closed in favour of https://github.com/pik-piam/mrdrivers/pull/86

PR #217, opened 2024-09-09
Bugfix: handle forcecache of length > 1

robinhasse:
`madrat:::cacheName` failed if `getConfig("forcecache")` had length > 1 and the `mode == "puc"`. I changed the respective condition in the if clause though I would ask you, @tscheypidi, to check if you now expect the intended behaviour.
I had the problem that if I force cache for specific functions by passing a vector of function names, the failure lead to no cache file being written which depending on the settings lead to a failure overall.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '62462280'
+ValidationKey: '62541725'
 AcceptedWarnings:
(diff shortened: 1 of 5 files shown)


tscheypidi (approved):
thanks for fixing it

PR #228, opened 2024-11-07
Fix for dev testthat

hadley:
`expect_true()` now applies a stricter test


pascal-sauer:
The tests would almost certainly break with these changes, so closing this. Seems like an automatic fix was applied that does not correctly recognize the commas used when subsetting a magclass object.

Issue #226, opened 2024-11-04
prepExtendedComment() breaks REMIND input data generation


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
See `/p/tmp/lavinia/MO/REMIND_pre-processing/pre-processing/log-2024-11-04-1388968.out`:
```
<callr_error/rlib_error_3_0/rlib_error/error>
Error: 
! in callr subprocess.
Caused by error in `get(as.character(cl[[1]]), mode = "function", sys.frame(-n - …`:
! object 'readSource' of mode 'function' was not found
```
```
Subprocess backtrace:
 1. base::do.call(madrat::retrieveData, c(cfg$args, list(renv = FALSE)))
 2. (function (model, rev = 0, dev = "", cachetype = "def", puc = identical(de…
 3. withr::with_dir(outputfolder, { …
 4. base::force(code)
 5. madrat:::withMadratLogging(do.call(cfg$functionCode, cfg$fullNow))
 6. base::withCallingHandlers(expr, message = messageHandler, warning = warningHandl…
 7. base::do.call(cfg$functionCode, cfg$fullNow)
 8. (function () …
 9. madrat::calcOutput(type = "Steel_Projections", subtype = "secondary.steel.max.shar… at fullREMIND.R:37:3
10. readSource(type = "ExpertGuess", subtype = "Chinese_Steel_Production", … at fullREMIND.R:37:3
11. quitte::madrat_mule(.)
12. madrat::readSource(type = "ExpertGuess", subtype = "Chinese_Steel_Production", … at madrat_mule.R:24:5
13. local .getData(type, subtype, subset, args, prefix)
14. madrat:::prepExtendedComment(xList, type, n = 2, warn = FALSE)
15. base::get(as.character(cl[[1]]), mode = "function", sys.frame(-n - …
16. base::.handleSimpleError(function (e) …
17. global h(simpleError(msg, call))
Error: retrieveData stopped due to an error. Search the logfile for 'Error'
Execution halted
```

LaviniaBaumstark:
@tscheypidi it is still an issue for REMIND input data generation. Could you or @pascal-sauer please have a look? Thanks

pascal-sauer:
I just realized the github action produces the error from the log when running my new test, so that is a hint :thinking:

pascal-sauer:
I can reproduce the error from the log by simply running `madrat:::testPrepExtendedComment()` with my new madrat version :+1:

[...]

PR #229, opened 2024-11-11
fix prepExtendedComment called from package::fun

pascal-sauer:


pascal-sauer:
fixes #226

R/prepExtendedComment.R
```diff
@@ -13,15 +13,24 @@
...
+  functionCall <- as.character(cl[[1]])
+
+  # if readSource is called as madrat::readSource functionName will
+  # be in this unintuitive order c("::", "madrat", "readSource")
+  if (length(functionCall) == 3 && functionCall[[1]] == "::") {
+    f <- get(functionCall[[3]], envir = loadNamespace(functionCall[[2]]), mode = "function", sys.frame(-n - 1))
+  } else {
+    f <- get(functionCall, mode = "function", sys.frame(-n - 1))
+  }
+
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 33: A bit unsafe to look for the function in some environment when `functionCall[[2]]` explicitly states it.  But this is an internal function, so OK.
But why not
```
functionCall <- tail(as.character(cl[[1]]), 1)
```
?

pascal-sauer: > A bit unsafe to look for the function in some environment when functionCall[[2]] explicitly states it

I assume you refer to the else branch? This is the behavior we had previously which I did not see a reason to change.

```
functionCall <- tail(as.character(cl[[1]]), 1)
```
I assume you suggest to skip the if else and use this instead to then call `get(functionCall, mode = "function", sys.frame(-n - 1))`? This won't work in a scenario where `madrat::readSource` is called, unless `madrat` is also attached (in which case readSource likely won't be called with the `madrat::` prefix), because `get` would only look on the search path.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > The `pos` argument can specify the environment in which to look for the object in any of several ways: as a positive integer (the position in the `search` list); as the character string name of an element in the search list; or as an `environment` (including using `sys.frame` to access the currently active function calls). The default of `-1` indicates the current environment of the call to `get`. The `envir` argument is an alternative way to specify an environment. 

—https://rdrr.io/r/base/get.html

The fully qualified call is
```
f <- get(functionCall[[3]], 
         envir = loadNamespace(functionCall[[2]]),
         mode = "function",
         pos = sys.frame(-n - 1))
```
Now, which argument is used, `envir` or `pos` (as they point to different things)?
<details>
  <summary>Spoiler:</summary>
  It is <code>envir</code>.
</details>

Since `prepExtendedComment()` is not exported, people calling it without attaching `madrat` are wholly responsible for their own pain `;)`.

But a safe formulation would be
```
f <- get(x = tail(functionCall, 1),
         envir = ifelse(1 == length(functionCall),
                        sys.frame(-n - 1),                # function() call
                        getNamespace(functionCall[[2]])), # pkg::function() call
         mode = "function")
```

PR #238, opened 2025-02-28
REUSE compliance

pascal-sauer:
One of the talks at deRSE was about the [command line tool](https://github.com/fsfe/reuse-tool) reuse` which helps with open source projects becoming [REUSE compliant](https://reuse.software/faq/). This makes sure that license information is explicit and clear for every single file, so the project is indeed safe (in a legal sense) to use as the license states. The tool can also check (e.g. in CI) whether a project is still REUSE compliant (`reuse lint`). This PR shows how that could look like for an R package, but there are other ways, e.g. name every single file or make sure all files are matched via globbing in the `REUSE.toml` file, instead of putting comment headers everywhere). This PR is meant as a basis for discussion if and how we want our projects to become REUSE compliant or not.

Related issue: [this article](https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code#why-not-bump-the-year-on-change) makes a case for not touching copyright years (as we do for magpiemodel), but instead just leave the year a file was created in the copyright notice.


pascal-sauer:
closed in favor of #248

Issue #243, opened 2025-03-13
Cache files could/shoud store the call that caused them to be written


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
That would help with debugging.  The calls in the `diagnostics.log` are often truncated.
Would also allow for checking of hash collisions 🤓

PR #246, opened 2025-03-17
callString attr for cache files, minor simplifications & clarifications

pascal-sauer:
I decided against removing the `...` feature for calc functions, as I did not find how that could simplify the code significantly. I simplified and clarified some other parts though, and implemented #243


R/cachePut.R
```diff
@@ -1,31 +1,28 @@
...
 #'
 #' @author Jan Philipp Dietrich
 #' @seealso \code{\link{cachePut}}, \code{\link{cacheName}}
 #' @examples
 #' \dontrun{
-#' example <- 1
-#' madrat:::cachePut(example, "calc", "Example", packages = "madrat")
+#' madrat:::cachePut(1, "calc", "Example", NULL, 'calcOutput("Example")')
 #' }
-cachePut <- function(x, prefix, type, args = NULL, graph = NULL, ...) {
```
tscheypidi, line 21: why did you remove the graph-argument? To my understanding that should quite significantly increase the runtime as the graph needs then to be recomputed every time

tscheypidi (changes requested):

R/cachePut.R
pascal-sauer: This argument is not used anywhere. To cope with the heavy complexity of madrat's internal functions I'd like to simplify as much as possible. If we decide to actually implement this performance improvement that you describe we can easily add this back in.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > To my understanding that should quite significantly increase the runtime as the graph needs then to be recomputed every time

No, because [it is cached](https://github.com/pik-piam/madrat/blob/f67bb979927ddc27ae8faf06fb1ee7a547e6a20b/R/getMadratGraph.R#L20-L34).

tscheypidi (approved):

PR #250, opened 2025-04-09
Update SIMSON countries

Merjo:
update 143 country mapping due to different spellings in SIMSON data and sort the list alphabetically (by country name/spelling, not ISO code)

### DIFF
--- a/inst/extdata/country2iso.csv
+++ b/inst/extdata/country2iso.csv
@@ -1,147 +1,162 @@
 ;x
 AD;AND
 AE;ARE
-Aegypten;EGY
-Aethiopien;ETH
 AF;AFG
-Afghanistan;AFG
 AG;ATG
-agypten;EGY
-Ägypten;EGY
 AI;AIA
 AL;ALB
-Aland Islands;ALA
-Åland Islands;ALA
-Albania;ALB
-albanien;ALB
-Algeria;DZA
-algerien;DZA
 AM;ARM
+AO;AGO
+AQ;ATA
+AR;ARG
+AS;ASM
+AT;AUT
+AU;AUS
+AUSTRALI;AUS
+AW;ABW
+AX;ALA
+AZ;AZE
+Aegypten;EGY
+Aethiopien;ETH
+Afghanistan;AFG
+Aland Islands;ALA
+Albania;ALB
+Algeria;DZA
 Amer Samoa;ASM
 American Samoa;ASM
 American_Samoa;ASM
 Andorra;AND
 Angola;AGO
+Angola (e);AGO
 Anguilla;AIA
-antarctic fisheries;XXX
 Antarctica;ATA
 Antigua & Barbuda;ATG
-Antigua and Barbuda;ATG
 Antigua Barb;ATG
-AO;AGO
-AQ;ATA
-aquatorialguinea;GNQ
-äquatorialguinea;GNQ
-AR;ARG
+Antigua and Barbuda;ATG
 Argentina;ARG
+Argentina (1);ARG
 Argentinien;ARG
 Armenia;ARM
 Armenien;ARM
 Aruba;ABW
-AS;ASM
-aserbaidschan;AZE
-AT;AUT
-athiopien;ETH
-Äthiopien;ETH
-AU;AUS
-AUSTRALI;AUS
 Australia;AUS
 Australien;AUS
 Austria;AUT
-AW;ABW
-AX;ALA
-AZ;AZE
+Austria (1);AUT
 Azerbaijan;AZE
 BA;BIH
-Bahamas;BHS
-Bahamas_The;BHS
-Bahrain;BHR
-bangladesch;BGD
-Bangladesh;BGD
-Barbados;BRB
 BB;BRB
 BD;BGD
 BE;BEL
-Belarus;BLR
-Belgien;BEL
-Belgium;BEL
-Belize;BLZ
-Benin;BEN
-Bermuda;BMU
 BES Islands;BES
 BF;BFA
 BG;BGR
 BH;BHR
-Bhutan;BTN
 BI;BDI
 BJ;BEN
 BL;BLM
 BM;BMU
 BN;BRN
 BO;BOL
+BOSNIAHERZ;BIH
+BQ;BES
+BR;BRA
+BS;BHS
+BT;BTN
+BV;BVT
+BW;BWA
+BY;BLR
+BZ;BLZ
+Bahamas;BHS
+Bahamas_The;BHS
+Bahrain;BHR
+Bangladesh;BGD
+Barbados;BRB
+Belarus;BLR
+Belgien;BEL
+Belgium;BEL
+Belize;BLZ
+Benin;BEN
+Bermuda;BMU
+Bhutan;BTN
 Bolivia;BOL
+Bolivia (3);BOL
 Bolivia (Plurinational State of);BOL
 Bolivia (Plurinational State of) ;BOL
 Bolivia, Plurinational State of;BOL
 Bolivien;BOL
-bonaire, saint eustatius, and saba;BES
 Bonaire, Sint Eustatius and Saba;BES
 Bosnia & Herzegovina;BIH
-Bosnia and Herzegovina;BIH
+Bosnia - Herzegovina;BIH
 Bosnia Herzg;BIH
-Bosnia_and_Herzegovina;BIH
+Bosnia and Herzegovina;BIH
 Bosnia-Herzegovina;BIH
-BOSNIAHERZ;BIH
+Bosnia_and_Herzegovina;BIH
 Bosnien und Herzegowina;BIH
 Botsuana;BWA
 Botswana;BWA
 Bouvet Island;BVT
-BQ;BES
-BR;BRA
 Br Virgin Is;VGB
+Br. Virgin Isds;VGB
 Brasilien;BRA
 Brazil;BRA
+Brazil (3);BRA
 British Indian Ocean Territory;IOT
 British Virgin Islands;VGB
 Brunei;BRN
 Brunei (darussalam);BRN
 Brunei Darsm;BRN
 Brunei Darussalam;BRN
-BS;BHS
-BT;BTN
 Bulgaria;BGR
-bulgarien;BGR
+Bulgaria ( from 1996 );BGR
+Bulgaria (1)(from 1996);BGR
 Burkina Faso;BFA
 Burkina_Faso;BFA
 Burma;MMR
 Burundi;BDI
-BV;BVT
-BW;BWA
-BY;BLR
 Byelarus;BLR
-BZ;BLZ
+Byelorussia;BLR
 CA;CAN
+CC;CCK
+CD;COD
+CF;CAF
+CG;COG
+CH;CHE
+CI;CIV
+CK;COK
+CL;CHL
+CM;CMR
+CN;CHN
+CO;COL
+CONGOREP;COD
+COTEIVOIRE;CIV
+CR;CRI
+CU;CUB
+CV;CPV
+CW;CUW
+CX;CXR
+CY;CYP
+CZ;CZE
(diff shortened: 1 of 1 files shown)


JakobBD:
Thanks for this!
- I could imagine @pascal-sauer would prefer the sorting to stay as it was, to be minimally intrusive? 
- I will open an issue for the error in the check

pascal-sauer (changes requested):
Seems like you were sorting part of the list alphabetically, just sort the whole list alphabetically and it's fine. Then please run `lucode2::buildLibrary()` and create a new minor version number, otherwise your changes won't become available when people update madrat :)

Merjo:
> I think @Merjo you sorted uppercase before lowercase, which is the Unicode/ASCII order. If you used python, you can try `sorted_list = sorted(unsorted_list, key=str.casefold)`

@JakobBD Yeah thanks, that's exactly what I'm working on right now, also seems like the previous sorting method ignores German Umlauts and Å and so forth, will update that soon, also thanks @pascal-sauer for the quick reply, will do so (I'm just slowly learning madrat :) )

Merjo:
@pascal-sauer I changed the sorting order now so should be better, but when I tried to run build library it didn't work.. I attached the log I got here, do you know what might be wrong? Sorry, as said a bit new to this so probably doing various amateur mistakes... :)

pascal-sauer:
Hi @Merjo a couple things are weird in the log you provided:
1) `Run lapply(list(...), function(x) if (is.list(x)) deparse(x) else x)`
This should instead be
`Run pucAggregate(...)`
which would let the test succeed. I have no idea why that is not the case.
2) Your madrat mainfolder does not seem to be set. I suggest to put something like `options(MADRAT_MAINFOLDER="C:/PIK/inputdata")` into your `.Rprofile` as described here https://gitlab.pik-potsdam.de/rse/rsewiki/-/wikis/Configuration-of-repositories-Add-the-PIK-repository
3) You have a puc folder in the madrat project folder. Could be related to 2. Try deleting this.

You can also ask the REMIND RSE folks for further help with your setup :)

inst/extdata/country2iso.csv
```diff
@@ -258,7 +293,14 @@ Eswatini;SWZ
 ET;ETH
 Ethiopia;ETH
 Ethiopia PDR;XET
+F. R. Germany;DEU
+F.R. Germany;DEU
+F.R. of Germany;DEU
+F.R. Yugoslavia;SCG
```
pascal-sauer, line 299:
```suggestion
F.R. Yugoslavia;YUG
```
Further below we also have
```
Yugoslavla;YUG
```

inst/extdata/country2iso.csv
```diff
@@ -311,12 +359,18 @@ GA;GAB
...
 Gabun;GAB
 Gambia;GMB
+Gambia, The;GMB
 Gambia_The;GMB
 GB;GBR
 GD;GRD
 GE;GEO
 Georgia;GEO
 Georgien;GEO
+German D.R.;DDR
```
pascal-sauer, line 369: We didn't have DDR in here before, and according to https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3 it is deleted. What are the implications if we add this here now?

inst/extdata/country2iso.csv
Merjo: The F.R. (Federal Republic of Yugoslavia) was a successor state to Yugoslavia as we know it and refers to only Serbia and Montenegro, with the (former) Iso code SCG, I was quite confused about it at first but at least in our data that is why.. though if this is risky with other projects I can also create an explicit mapping for that in our functions (s. https://en.wikipedia.org/wiki/Serbia_and_Montenegro)

pascal-sauer: Ah okay, in that case all good :)

pascal-sauer (approved):

inst/extdata/country2iso.csv
JakobBD: There never was `GDR` in any ISO 3166 code, right? 
So it seems like a bug that should be sorted out? 
Should I open an issue? 

(We currently seem to use the former ISO 3166-1 3-letter codes for historical regions. 
This seems a reasonable simplification compared to the "clean" solution of using ISO 3166-3 4-letter-codes. 
But it may be good to make this explicit as a rule, which we follow and document in the descriptions of `toolISOhistorical` and `toolCountry2isocode`, to avoid such confusion in the future?)

LaviniaBaumstark: I also agree, DDR is at least a deleted ISO3 country code. Could you please check that toolISOhistorical can deal witht it?

Merjo: I just tested it with toolISOhistorical and it does not work (at least with our steel production data) because in the ISOhistorical.csv, East Germany is defined as GDR. I could fix it by replacing the lines

GDR;DEU;y1990
BRG;DEU;y1990

with

DDR;DEU;y1990
DEU;DEU;y1990

Just adding those to lines at the end does not work as toolISOhistorical then still expects there to be some 'BRG' and 'GDR' data. If we just mapped East Germany to GDR in this file (country2iso.csv), it still does not work because it expects 'BRG' data. However, we can't map Germany just to 'BRG' as then in other contexts it has to be the official code DEU. I feel like the options we have are:

1. Change the ISOhistorical.csv as I did above, which seems sensible to me as West Germany's ISO code was DEU and never BRG, but not sure if this will cause problems in any other data preparation?

2. Leave it as it is and therefore leave East Germany out of this file (country2iso.csv), and then I will add a specific mapping when loading toolCountry2isocode.

I will add the proposed changes now along with some additional mappings I missed in a new commit now so everyone can see.

What do you all think, maybe there are better alternatives? Personally, I'm happy with either option, out of a steel data perspective the first approach is more elegant and maybe also helpful with other datasets but I can not really assess how much that might destroy other data loading routines.

Sorry for the long text and thanks everyone for your help!! :)

JakobBD: Thanks for all this effort!
I'm in favor of option 1. If the others agree, don't forget to also change the `XDE;...;y1944` lines and the entries in `country2iso.csv` accordingly. (`XDE` itself seems also made up and I would leave it like that.)

Could we then leave out `DEU;DEU;y1990` altogether? Not sure about the syntax...

I did a ripgrep on the source folder on the cluster and there is neither GDR, nor the according country name in `country2iso.csv`. So chances are we could be fine.

pascal-sauer (changes requested):

inst/extdata/country2iso.csv
Merjo: Thanks for the correction of my mistake @pascal-sauer !

Merjo: I just checked again and realized a mistake: The way I proposed led to Germany data being deleted altogether. That was due to a part of toolISOhistorical that deleted the old countries but when aggregating countries where one country exists before and after the aggregation (here 'DEU') it is deleted as well. 

Additionally, when not selecting overwrite=TRUE, there was of course data of the new country in the past, and hence if that can not be overwritten nothing happens. I changed the code in toolISOhistorical accordingly (basically just moving some statements around so it reflects the aggregation case better). It's added in a new commit.

So again we could either (1.) leave it as it was and have a special routine for the steel data or (2.) use my new changes, hoping it will not break anything. 

Thanks for the help so far, also @JakobBD for the ripgrep!

Merjo:
@pascal-sauer @JakobBD 

So I realized that my proposed option the other day did not work because it deleted Germany data pre 1990 altogether. I wrote this the other day but forgot to tag you specifically about that: It "was due to a part of toolISOhistorical that deleted the old countries but when aggregating countries where one country exists before and after the aggregation (here 'DEU') it is deleted as well.

Additionally, when not selecting overwrite=TRUE, there was of course data of the new country in the past, and hence if that can not be overwritten nothing happens. I changed the code in toolISOhistorical accordingly (basically just moving some statements around so it reflects the aggregation case better)."

The change is in a new commit and I just added an additional commit because this led to problems with the test function of toolISOhistorical. I realize that this is a lot of changes to a crucial function that might lead to bigger problems elsewhere (though I hope it potentially could improve other code as well?). We could either just leave it as it was before and I create a separate function so that the historical (dis)aggregation is done as I need it for SIMSON or use my new changes with the danger of creating further problems... what do you think is the best option?

JakobBD:
Thanks for the effort @Merjo! To me, it seems to make sense, but then again, I don't have much of an actual clue. 
Looking forward to @pascal-sauer s opinion!

R/toolISOhistorical.R
```diff
@@ -152,25 +152,37 @@ toolISOhistorical <- function(m, mapping = NULL, additional_mapping = NULL, over
...
+        # fill data
+        if (overwrite) {
+          m[a$toISO, subTime, ] <- mTr
+        } else {
+          m[a$toISO, subTime, ][is.na(m[a$toISO, subTime, ])] <- mTr[is.na(m[a$toISO, subTime, ])]
+        }
+        
+        # delete old lines
+        for (b in a$fromISO) {
+          if (is.element(b, getItems(m, dim = 1.1))) {
```
pascal-sauer, line 165: I find it a bit easier to read this way, just an option:
```suggestion
          if (b %in% getItems(m, dim = 1.1)) {
```

R/toolISOhistorical.R
```diff
@@ -152,25 +152,37 @@ toolISOhistorical <- function(m, mapping = NULL, additional_mapping = NULL, over
...
-      if (overwrite) {
+        
+        # fill data, here no option to not overwrite old data as countries are aggregated
         m[a$toISO, subTime, ] <- mTr
-      } else {
-        m[a$toISO, subTime, ][is.na(m[a$toISO, subTime, ])] <- mTr[is.na(m[a$toISO, subTime, ])]
+        
+        # delete old lines
+        for (b in a$fromISO) {
+          if ((is.element(b, getItems(m, dim = 1.1))) & (!(is.element(b, a$toISO)))) { # only delete lines that are not resulting countries
```
pascal-sauer, line 179: Did you get a linter warning here? In if statements `&&` should be used over `&`, as `&&` always results in a single TRUE/FALSE, while `&` can create a vector of TRUE/FALSE which does not work for `if`.
```suggestion
          if ((is.element(b, getItems(m, dim = 1.1))) && (!(is.element(b, a$toISO)))) { # only delete lines that are not resulting countries
```

pascal-sauer (commented):
Thanks, these changes make sense to me! I wonder if we were accidentally removing `ANT` before, because that we also mapped onto itself in toolISOhistorical: `ANT;ANT;y1985` :sweat_smile:

pascal-sauer:
To make sure this does not break existing code I'd suggest you run the full remind preprocessing with the changes you propose here with the help of a remind RSE person. If that's all good we can merge I'd say :)

pascal-sauer:
And before that please merge upstream changes into your fork/branch and run buildLibrary again

R/toolISOhistorical.R
Merjo: Ah, I did not know that syntax, sounds good!

R/toolISOhistorical.R
Merjo: I actually did not for some reason, though that sounds good to me. Maybe it makes sense to use the 'in' syntax as above here as well?

fbenke-pik:
Update: this PR will be closed as we opted for cleansing the data in mrmfa rather than delegating special cases to `toolISOhistorical` and enhancing the logic of `toolISOhistorical`. 

I don't want to merge the enhancements of  `country2iso.csv` either, as these include a lot of very specific names that should rather be tackled by data cleansing in mrmfa, e.g. `Austria (1);AUT`

See also https://github.com/bennet21/mrmfa/pull/2

However, I might do a follow-up PR at some point with a few more universal additions to `country2iso.csv`  that are currently mapped in mrmfa.

PR #253, opened 2025-07-01
Deprecates setting the diagnostics config option to a filename.

codeZeilen:
There are no known uses throughout Github repos (including pik-piam).

One disadvantage is now that we hard code the file name between `vcat` and `retrieveData`: 

https://github.com/pik-piam/madrat/blob/c6f192af5908dc77f1067b023035a2fa14d010ba/R/retrieveData.R#L185

https://github.com/pik-piam/madrat/blob/c6f192af5908dc77f1067b023035a2fa14d010ba/R/vcat.R#L75

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '64663401'
+ValidationKey: '65066700'
 AcceptedWarnings:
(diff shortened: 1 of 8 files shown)


tscheypidi (approved):
looks good

PR #255, opened 2025-07-10
Adds toolCompareStatusLogs to get a diff between two status logs

codeZeilen:
This PR adds the `toolCompareStatusLogs` function that compares two status logs (either from files or full tgz archives) and returns a textual rendering of a diff describing the changes. The diff is supposed to be human-readable and more  of a report, than a full diff format. An alternative representation could be generated from the return value of `.compareStatusLogsStatistics`.

Currently, the comparison treats any structured entries other than statistic entries as values, thus any change in the nested entry results in the old version being recorded as removed and the new one being recorded as added.

This PR only includes:
- The added function `toolCompareStatusLogs` and auxiliary functions
- Corresponding tests

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '66239125'
+ValidationKey: '66451840'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -2,8 +2,8 @@ cff-version: 1.2.0
 message: If you use this software, please cite it using the metadata from this file.
 type: software
 title: 'madrat: May All Data be Reproducible and Transparent (MADRaT) *'
-version: 3.25.1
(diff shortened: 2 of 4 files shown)


R/toolCompareStatusLogs.R
```diff
@@ -0,0 +1,222 @@
...
+
+.compareStatusLogsStatistics <- function(oldLog, newLog) {
+  # Assumed structure: list(string = call(list(entry)), entry = string|list
+
+  # Detect added/removed calls
+  removedCalls <- as.list(setdiff(names(oldLog), names(newLog)))
+  addedCalls <- as.list(setdiff(names(newLog), names(oldLog)))
+
+  # Detect changed calls
+  sameCalls <- as.list(intersect(names(newLog), names(oldLog)))
```
codeZeilen, line 63: @pascal-sauer Do you have a better solution for this pattern (L63-68)? I have used this several times now. It is basically a map that expands a vector of tokens into a named list with the tokens as keys and a values derived from the tokens as the values.

pascal-sauer:
Looks good to me, some minor things should be double checked.

PR #263, opened 2025-09-16
fix bug related to file parameter in calcOutput

fbenke-pik:
Currently, file names containing `mif` are matched as `*.mif`, e.g. `emi2020_sectGAINS_sourceCEDS.cs4r` produces a mif file instead of a cs4r file.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '65901600'
+ValidationKey: '65944627'
 AcceptedWarnings:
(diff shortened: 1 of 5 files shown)


pascal-sauer:
Is `emi2020_sectGAINS_sourceCEDS.cs4r` really the filename you wanted to put here? It does not match `.mif`, does it?

laurinks:
@fbenke-pik wanted to put `emifacs_sectGAINS_sourceCEDS.cs4r` :)

PR #302, opened 2026-06-01
Fix for retrieve data permissions test

codeZeilen:


codeZeilen:
@pascal-sauer Changes look wild, as I intentionally included the commits from the other PR

codeZeilen:
I have changed the actual assertions to only check the user permissions, as the `umask` may be different for different environments, but I would argue that it is save to assume that we get write permissions for the user.

pascal-sauer (approved):

PR #308, opened 2026-07-14
fix toolFixWeight

pascal-sauer:
- adds toolFixWeight support for 3 or more subdims (X.Y.Z)
- import cleanup to fix linter warning


tscheypidi (approved):

codeZeilen (approved):
Took me a while to trace what is happening, but looks good to me now.

pascal-sauer:
@codeZeilen is that because of the new code, or because toolAggregate is very complex? I have trouble navigating toolAggregate with all it's sub functions too :grimacing:

PR #309, opened 2026-07-14
Add neasted calling of `withMadratLogging` with different parameters and add `warningsAsErrors` parameter

tobiasdiez:
This PR changes how withMadratLogging() behaves when calls are nested.

Previously, the first, outermost call installed the handlers. Every nested withMadratLogging() call then returned its expression immediately. As a result, its `logOnly` argument had no effect: all warnings and errors were handled according to the outermost call’s settings.

With this PR, each withMadratLogging() call temporarily introduces its own policy with the following update mechanism:
- An explicitly supplied nested argument overrides settings locally.
- An omitted nested argument inherits the enclosing policy.

Also this PR introduces a `warningsAsErrors` argument to `withMadratLogging` so that warnings are logged as errors with the correct verbosity handling.

### DIFF
--- a/R/downloadSource.R
+++ b/R/downloadSource.R
@@ -122,9 +122,7 @@ downloadSource <- function(type, subtype = NULL, overwrite = FALSE, numberOfTrie
   with_dir(downloadInProgressDirectory, {
     setWrapperActive("wrapperChecks")
     # convert warnings to errors, because in download functions warnings usually mean data was not downloaded properly
-    withr::with_options(c(warn = 2), {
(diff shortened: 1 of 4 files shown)


tobiasdiez:
Running `buildLibrary` hangs for me with:
```
....
─  checking tests ... SKIPPED
✔  checking for unstated dependencies in vignettes ...
✔  checking package vignettes ...
✔  checking re-building of vignette outputs (3.9s)
✔  checking for non-standard things in the check directory
✔  checking for detritus in the temp directory ...
   
```

any idea what's the problem?

PR #312, opened 2026-07-31
Flexible cache format for madrat

codeZeilen:
This PR adds the feature to configure the file format used in caches to allow us to use qs2 which makes use of the cores we use in jobs to speed up writing (~5-10x) and reading (~2-3x).

The main change is that one can now register a cache format and then select one through a config or env variable. The cache format will fall back to rds reading if there is no cache file available in the desired format yet. 

The second change that is necessary is the conversion for generating PUCs at the end. For that all cache files to be included in the PUC are converted back to rds. To speed this up the cache format can provide a direct format->rds conversion function (qs2 for instance supports this).

Tests needed a lot of adjustments as they relied on `.rds` as extension.

TODO:
- [x] Check with CRAN workflows 
- [x] CRAN wincheck
- [x] Decide whether we want to keep the qs2 suggest
- [x] Evaluate correctness and runtime on full vanilla preprocessing
- [x] Evaluate correctness and runtime on full qs2 preprocessing
- [x] Check with REMIND preprocessing

Disclaimer: Large parts are generated using Claude then reviewed and adjusted.


R/cacheFormat.R
```diff
@@ -0,0 +1,144 @@
...
+  return(format)
+}
+
+# file extensions to look for when searching a cache file, in order of preference:
+# the configured format first, rds last as it is always readable (see cacheNames)
+cacheExtensions <- function() {
+  return(unique(c(cacheFormat()$extension, "rds")))
+}
+
+#' @importFrom tools file_ext
```
pascal-sauer, line 100: A bit weird to import here, instead of below where it's used, but it works 🤷

pascal-sauer: Did you also find out again that we cannot use tools::file_ext as madrat misidentifies this as a tool function call? 🙈

pascal-sauer: If you and/or your claude companion are ever bored... Cannot be too hard to fix 😉

R/cacheNames.R
```diff
@@ -0,0 +1,96 @@
+#' cacheNames
```
pascal-sauer, line 1: Was this an actual rewrite, or was cacheName.R just renamed? In case of the latter, could you tell git that it's actually a rename, so changes are displayed properly? Makes the review easier. I'll not look at this file for now

R/cachePut.R
```diff
@@ -40,14 +45,22 @@ cachePut <- function(x, prefix, type, fname, callString) {
     attr(x, "madratMessage") <- getMadratMessage(fname = paste0(prefix, type))
     attr(x, "callString") <- callString
 
-    # write to tempfile to avoid corrupt cache files in parallel running preprocessings
-    tempfileName <- paste0(fname, Sys.getenv("SLURM_JOB_ID", unset = ""))
-    saveRDS(x, file = tempfileName, compress = getConfig("cachecompression"))
-    file.rename(tempfileName, fname)
+    # write to tempfile to avoid corrupt cache files in parallel running preprocessings. The
+    # leading "." keeps the tempfile out of the Sys.glob cache file search in cacheNames.
```
pascal-sauer: Oh, this might be super relevant, no? 💡

R/initializeConfig.R
```diff
@@ -35,6 +35,7 @@ initializeConfig <- function(verbose = TRUE) {
                 forcecache           = FALSE,
                 ignorecache          = NULL,
                 cachecompression     = "gzip",
+                cacheformat          = Sys.getenv("MADRAT_CACHEFORMAT", unset = "rds"),
```
pascal-sauer, line 38: So we need MADRAT_CACHEFORMAT=qs2 in the piam apptainer/module?

R/pucAggregate.R
```diff
@@ -43,14 +43,17 @@ pucAggregate <- function(puc, regionmapping = getConfig("regionmapping"), ..., r
...
+                            madratCacheFormats) {
     # need to use `::` because this is run in another R session
     if (file.exists("puc/renv.lock")) {
       renv::init()
       renv::restore(lockfile = "puc/renv.lock", prompt = FALSE)
     }
     withr::local_options(madrat_cfg = madratCfg,
                          madrat_codelabels = madratCodelabels,
+                         # custom cache formats are not registered in this new session
+                         madrat_cacheformats = madratCacheFormats,
```
pascal-sauer, line 57: I was expecting `madrat_cacheformats = builtinformats[["rds"]]` or something like that, why is it like this?

R/retrieveData.R
```diff
@@ -380,6 +382,28 @@ retrieveData <- function(model, rev = 0, dev = "", cachetype = "def", puc = iden
   return(invisible(NULL))
 }
 
+#' .cacheFilesToRds
+#' Copies the given cache files into the current working directory, converting everything
+#' which is not already rds. Conversion failures are reported but do not raise an error, so
+#' that a lengthy data processing is never aborted by a problem with the puc file.
+#' @param cacheFiles Paths of the cache files to be put into a puc file.
+#' @return The file names as they should now exist in the current working directory.
+#' @importFrom tools file_path_sans_ext file_ext
```
pascal-sauer:
```suggestion
#' @importFrom tools file_path_sans_ext
```

R/setConfig.R
```diff
@@ -53,7 +53,13 @@
 #'  The top level function must always be part of this list.
 #' @param cachecompression logical or character string specifying whether cache files
 #' use compression. TRUE corresponds to gzip compression, and character strings "gzip",
-#' "bzip2" or "xz" specify the type of compression.
+#' "bzip2" or "xz" specify the type of compression. Only applies to the cache format "rds".
+#' @param cacheformat name of the serialization format used for cache files. Default is
+#' "rds", madrat also ships "qs2" (which is considerably faster, but requires the \code{qs2}
+#' package). Further formats can be added via \code{\link{registerCacheFormat}}. Cache files
+#' are written in the configured format, but an already existing rds file is still read if no
```
pascal-sauer, line 60: We are only falling back to rds, but never to qs2, yes? Hypothetically, if we switch to qs3 or so, it might make sense to use rds/qs2 opportunistically as fallback. Not sure if it's worth the effort to change now, but in case this is just inconsistent documentation it could be fixed.

tests/testthat/test-cacheCopy.R
```diff
@@ -1,17 +1,18 @@
 log <- c("Current madrat configuration:",
```
pascal-sauer: You're fixing a linter warning here, no? Shouldn't it be like this?
```suggestion
log <- c(
  "Current madrat configuration:",
```

pascal-sauer: If there's no linebreak after `(` I thought the linter would require indentation and braces like this:
`log <- c("asd",
          "fgh")`
Is that not the case?

tests/testthat/test-cacheFormat.R
```diff
@@ -0,0 +1,299 @@
+# NOTE: the calc functions defined in the tests below must all have distinct bodies. The
+# madrat graph is cached (see getMadratGraph) under a hash which does not distinguish
+# identically defined functions, so reusing a body makes a later test look up a function
+# which is not part of the cached graph.
```
pascal-sauer, line 4: That's good to keep in mind 💡 
Note: I'm not reviewing this 300 lines test file in detail, I trust you carefully checked it tests what it should :)

pascal-sauer (commented):
Thank you, great stuff! Will have a look at cacheNames once we clarified if it's a rewrite or file rename.

R/cacheNames.R
codeZeilen: Hm... I think this reaches the limits of git change tracking. It was marked as a rename but also includes a lot of changes, so I think it reached the threshold for being marked as a new file. 

Here is the manual diff between cacheName.R and cacheNames.R (looks worse than it is, some blocks are just moved around):

```diff
diff --git a/home/patrickr/tmp/cacheName.R b/R/cacheNames.R
index 3865908d..e3e0a010 100644
--- a/home/patrickr/tmp/cacheName.R
+++ b/R/cacheNames.R
@@ -1,22 +1,32 @@
-#' cacheName
+#' cacheNames
 #'
-#' Get the name of a cache file corresponding to the given args
+#' Get the names of the cache files corresponding to the given args
 #'
-#' @note With \code{setConfig(forcecache=TRUE)} cacheName will also return
-#' cache file names with deviating fingerprint if no fitting cache file is found
-#' (if there are multiple it will just return the newest one).
+#' @note With \code{setConfig(forcecache=TRUE)} cacheNames will also look for
+#' cache files with deviating fingerprint if no fitting cache file is found
+#' (if there are multiple it will just use the newest one).
+#'
+#' Cache files are searched for in the configured cache format (see
+#' \code{\link{setConfig}}) first and in "rds" second. This way an existing rds
+#' cache stays usable after switching to another cache format. Files found via
+#' this fallback are only read, never rewritten, so the returned write target
+#' always uses the configured format.
 #'
 #' @param prefix function prefix (e.g. "calc" or "read")
 #' @param type output type (e.g. "TauTotal")
 #' @param args a list of named arguments used to call the given function
-#' @return absolute path to a cache file that does not necessarily exist
+#' @return a list with two elements: \code{write}, the absolute path to the cache
+#' file that should be written for the given arguments (in the configured cache
+#' format), which does not necessarily exist, and \code{read}, the absolute path
+#' of an already existing, fitting cache file which should be read (possibly in
+#' another format), or NULL if there is none.
 #'
 #' @author Jan Philipp Dietrich, Pascal Sauer
 #' @seealso \code{\link{cachePut}}
 #' @keywords internal
 #' @examples
-#' madrat:::cacheName("calc", "TauTotal")
-cacheName <- function(prefix, type, args = NULL) {
+#' madrat:::cacheNames("calc", "TauTotal")
+cacheNames <- function(prefix, type, args = NULL) {
   fpprefix <- prefix
   if (prefix %in% c("convert", "correct")) {
     fpprefix <- "read"
@@ -35,50 +45,52 @@ cacheName <- function(prefix, type, args = NULL) {
   argsHash <- cacheArgumentsHash(call, args,
                                  errorOnMismatch = !(prefix %in% c("read", "correct")))
 
-  .fname <- function(prefix, type, fp, argsHash) {
-    return(paste0(getConfig("cachefolder"), "/", prefix, type, fp, argsHash, ".rds"))
+  # vectorized over extension
+  .fname <- function(fp, argsHash, extension) {
+    return(paste0(getConfig("cachefolder"), "/", prefix, type, fp, argsHash, ".", extension))
   }
 
-  fname <- .fname(prefix, type, paste0("-F", fp), argsHash)
-  fnameNoFingerprint <- .fname(prefix, type, "", argsHash)
+  extensions <- cacheExtensions() # Contains configured extension as well as rds
+  forcecache <- isConfigSet(prefix, type, "forcecache")
 
-  if (file.exists(fname)) {
-    return(fname)
-  }
-  if (!isConfigSet(prefix, type, "forcecache")) {
-    vcat(2, " - Cache file ", basename(fname), " does not exist", show_prefix = FALSE)
-    # returning fname even though that file does not exist -> calling function must
-    # check whether the returned file exist
-    return(fname)
-  }
+  # The file to write depends only on forcecache and always uses the configured format.
+  writeName <- .fname(if (forcecache) "" else paste0("-F", fp), argsHash, extensions[1])
 
-  if (isConfigSet(prefix, type, "ignorecache")) {
-    # the returned path won't be read, but it's used for writing
+  # The file to read may be in any registered format, preferring the configured one.
+  readName <- NULL
+  fittingFiles <- .fname(paste0("-F", fp), argsHash, extensions)
+  existingFiles <- fittingFiles[file.exists(fittingFiles)]
+  if (length(existingFiles) > 0) {
+    # identical fingerprint means identical content, so prefer the configured format
+    readName <- existingFiles[1]
+  } else if (!forcecache) {
+    vcat(2, " - Cache file ", basename(fittingFiles[1]), " does not exist", show_prefix = FALSE)
+  } else if (isConfigSet(prefix, type, "ignorecache")) {
     vcat(2, " - forcecache and ignorecache are both active", show_prefix = FALSE)
-    return(fnameNoFingerprint)
-  }
+  } else { # forceCache is TRUE
+    # no perfectly fitting file exists, try to find a similar one for forcecache
+    # (either with no fingerprint hash or with differing fingerprint)
+    files <- Sys.glob(c(.fname("-F*", argsHash, extensions),
+                        .fname("", argsHash, extensions)))
 
-  # no perfectly fitting file exists, try to find a similar one for forcecache
-  # (either with no fingerprint hash or with differing fingerprint)
-  files <- Sys.glob(c(.fname(prefix, type, "-F*", argsHash),
-                      fnameNoFingerprint))
+    # remove false positives
+    if (is.null(argsHash)) {
+      files <- files[!grepl("-[^F].*$", basename(files))]
+    }
 
-  # remove false positives
-  if (is.null(argsHash)) {
-    files <- files[!grepl("-[^F].*$", basename(files))]
+    if (length(files) == 0) {
+      vcat(2, " - No fitting cache file available", show_prefix = FALSE)
+      vcat(3, " - Search pattern ", paste(basename(.fname("-F*", argsHash, extensions)), collapse = ", "),
+           show_prefix = FALSE)
+    } else {
+      # found one or more similar files, use the newest one
+      readName <- files[robustOrder(paste(file.mtime(files), basename(files)), decreasing = TRUE)][1]
+      if (!isWrapperActive("pucAggregate")) {
+        vcat(1, " - forced cache does not match fingerprint ", fp,
+             fill = 300, show_prefix = FALSE)
+      }
+    }
   }
 
-  if (length(files) == 0) {
-    vcat(2, " - No fitting cache file available", show_prefix = FALSE)
-    vcat(3, " - Search pattern ", basename(.fname(prefix, type, "-F*", argsHash)), show_prefix = FALSE)
-    return(fnameNoFingerprint)
-  } else {
-    # found one or more similar files, use the newest one
-    fname <- files[robustOrder(paste(file.mtime(files), basename(files)), decreasing = TRUE)][1]
-  }
-  if (!isWrapperActive("pucAggregate")) {
-    vcat(1, " - forced cache does not match fingerprint ", fp,
-         fill = 300, show_prefix = FALSE)
-  }
-  return(fname)
+  return(list(write = writeName, read = readName))
 }
 ```

File: R/cacheNames.R [hunk 10]
```diff
@@ -0,0 +1,96 @@
...
+#' @param prefix function prefix (e.g. "calc" or "read")
+#' @param type output type (e.g. "TauTotal")
+#' @param args a list of named arguments used to call the given function
+#' @return a list with two elements: \code{write}, the absolute path to the cache
+#' file that should be written for the given arguments (in the configured cache
+#' format), which does not necessarily exist, and \code{read}, the absolute path
+#' of an already existing, fitting cache file which should be read (possibly in
+#' another format), or NULL if there is none.
+#'
+#' @author Jan Philipp Dietrich, Pascal Sauer
```
pascal-sauer:
```suggestion
#' @author Patrick Rein, Jan Philipp Dietrich, Pascal Sauer
```

R/cacheNames.R
```diff
@@ -0,0 +1,96 @@
...
+                  paste0(prefix, type),
+                  attr(fp, "call"),
+                  fixed = TRUE))
+  }
+  argsHash <- cacheArgumentsHash(call, args,
+                                 errorOnMismatch = !(prefix %in% c("read", "correct")))
+
+  # vectorized over extension
+  .fname <- function(fp, argsHash, extension) {
+    return(paste0(getConfig("cachefolder"), "/", prefix, type, fp, argsHash, ".", extension))
```
pascal-sauer, line 50: Sidenote: Jan intentionally writes inner functions as if they do not have access to the outer scope. I prefer less args/code, so using vars from the outer scope is fine by me.

R/cacheNames.R
pascal-sauer: Thanks, I reviewed it now.

R/cachePut.R
codeZeilen: Yes, the whole point of this change is that the file-extension now becomes important. The tmp-file before had the SLURM_JOB_ID at the end which prevents cacheWrite from determining what to do :| Instead of this route we could also separate the actual fileName and the tempfileName and pass both to cacheWrite like so: `cacheWrite(x, file = fname, tmpFile = tempfileName)`. Then we could continue using the old rename.

OR: We could actually move this whole logic into `cacheWrite`, letting the "caching I/O" layer handle the renaming and permission setting (now that I write this, I think this is the better solution actually).

Any thoughts? :)

R/initializeConfig.R
codeZeilen: Yes, this + a setting of the qs2 `nthreads` option in the system .Rprofile.

As an alternative, we could only do this for the preprocessing: https://github.com/magpiemodel/preprocessing-magpie/pull/5/changes

So far, we always went with an apptainer/module setting, right?

R/pucAggregate.R
codeZeilen: Interestingly, the PUC aggregation needs the configured format ("qs2" for us), not "rds". For the conversion we need to read the cache files which were written in the configured format (or rds as fallback, thus all configured formats) and only then write as rds, regardless of what is configured. Also: The configured extension also may contain a fast-path conversion to rds.

R/initializeConfig.R
pascal-sauer: I'd argue we want it also for development purposes outside of the preprocessing repo. Yes, we kept these kind of things in the apptainer/module so it's simply all in one place (more or less)

R/cachePut.R
pascal-sauer: Moving this into cacheWrite sounds good and makes sense to me. No strong feelings here

pascal-sauer (approved):

PR #314, opened 2026-08-05
Resolving flakyness in PUC locking and toolFixWeight tests

codeZeilen:
This PR should put an end to the flaky PUC thread tests. Mostly by improving the log handling and the way processes are controlled form the outside.

The `processLog` helper is basically the old log object reworked to handle process termination better.

### DIFF
--- a/tests/testthat/setup.R
+++ b/tests/testthat/setup.R
@@ -36,6 +36,42 @@ globalassign <- function(...) {
   }
 }
 
+# Accumulating stdout reader for a callr background process. read_output_lines() is destructive, so
+# the lines have to be collected. waitFor fails if the process dies instead of waiting forever.
+processLog <- function(process, timeout = 60) { # timeout in seconds
+  log <- character(0)
+  update <- function() {
+    log <<- c(log, process$read_output_lines())
+    log
+  }
+  list(
+    waitFor = function(message) {
+      deadline <- Sys.time() + timeout
+      repeat {
+        if (message %in% update()) {
+          return(invisible(log))
+        }
+        if (!process$is_alive()) {
+          # read once more, the process may have printed the message right before exiting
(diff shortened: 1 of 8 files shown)


tests/testthat/setup.R
```diff
@@ -36,6 +36,42 @@ globalassign <- function(...) {
   }
 }
 
+# Accumulating stdout reader for a callr background process. read_output_lines() is destructive, so
+# the lines have to be collected. waitFor fails if the process dies instead of waiting forever.
+processLog <- function(process, timeout = 60) { # timeout in seconds
+  log <- character(0)
+  update <- function() {
+    log <<- c(log, process$read_output_lines())
```
pascal-sauer, line 44: Is `<<-` necessary? Is this ok for the linter?

pascal-sauer (approved):

tests/testthat/setup.R
codeZeilen: Funny enough the linter didn't care o.O

Yes, I originally also wrote it like that because the log "object" coming out of this `processLog` function uses the local variable `log` in the env created in this "constructor" as the actual shared state between the "methods". Writing with `<-` would only write into the env of the update function invocation.

It's somewhat the price for encapsulating the three functions into one entity. :|

pascal-sauer:
checks on windows on r-universe are [failing](https://github.com/r-universe/pik-piam/actions/runs/29510967202/job/87665164063), because toolFixWeight is too slow (but only on windows :grimacing: ). I'd just "fix" by allowing up to 10 seconds. If you agree this is okay, could you include it in this PR? Or do you think this should be a separate PR?

codeZeilen:
@pascal-sauer Sounds good to include. Then this is the fix flaky tests PR :)

PR #315, opened 2026-08-06
Improve performance of toolFixWeight slightly

codeZeilen:
This PR improves performance by about 1.7x, mostly through working on the data.frame columns individually instead of the complete data.frame.

### DIFF
--- a/R/toolFixWeight.R
+++ b/R/toolFixWeight.R
@@ -43,7 +43,7 @@
 toolFixWeight <- function(weight, map, dim) {
   dim <- dimCode(dim, weight)
   stopifnot(length(dim) == 1,
-            !any(weight < 0),
+            all(weight >= 0),
             ncol(map) == 2)
   if (!setequal(map[[2]], getItems(weight, dim))) {
     map <- map[, 2:1]
--- a/R/toolFixWeight.R
+++ b/R/toolFixWeight.R
@@ -43,7 +43,7 @@
 toolFixWeight <- function(weight, map, dim) {
   dim <- dimCode(dim, weight)
   stopifnot(length(dim) == 1,
-            weight >= 0,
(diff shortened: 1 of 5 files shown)


R/toolFixWeight.R
```diff
@@ -43,7 +43,7 @@
 toolFixWeight <- function(weight, map, dim) {
   dim <- dimCode(dim, weight)
   stopifnot(length(dim) == 1,
-            weight >= 0,
+            all(weight >= 0),
```
pascal-sauer, line 46: Does this impact performance? :O If not, I thought about this back then and decided against the `all` in `stopifnot` calls, it's just less code and slightly easier to read to me. No strong feelings though, we can change it if you want.

pascal-sauer (approved):

pascal-sauer:
How did you find this? I presume an agent found it, with a skill?

R/toolFixWeight.R
codeZeilen: It does (not much in absolute terms 2s -> 0.1s, but still)! 

I was also quite confused at first. The reason is that `weight >= 0` returns a full logical magpie object, on which `stopifnot` then does an `anyNA`, which takes basically two seconds (maybe also a problem). By doing `all` beforehand, we get the same `NA` handling in the end. The interaction between `any` and `all` and `stopifnot` works out pretty well.

codeZeilen:
Yes, simply plain Opus. Decided to briefly look into it after increasing the timeout of the test.

R/toolFixWeight.R
pascal-sauer: Okay, that's surprising, and good to know!

codeZeilen:
I am postponing merging this until the LPJmL update preprocessing has finally completed successfully.

PR #317, opened 2026-08-17
Adds memory profiling

codeZeilen:
This PR adds memory profiling based on OS-recorded memory usage (`/proc/self/status`) which is more useful for debugging OOM issues, as it reflects the OS-level memory usage.

The parsing of these logs was moved to another PR codezeilen/madrat#3


R/initializeConfig.R
```diff
@@ -39,6 +39,7 @@ initializeConfig <- function(verbose = TRUE) {
                 hash                 = "xxhash32",
                 diagnostics          = FALSE,
                 debug                = FALSE,
+                memoryProfiling      = FALSE,
```
pascal-sauer, line 42: What would be the performance implications of enabling this by default?

R/memoryProfiling.R
```diff
@@ -0,0 +1,74 @@
...
+  return(file.exists("/proc/self/status") && # nolint: absolute_path_linter.
+           file.exists("/proc/self/clear_refs")) # nolint: absolute_path_linter.
+}
+
+#' getProcMemory
+#'
+#' Reads the current and the peak resident set size of this process as recorded
+#' by the kernel.
+#'
+#' @return A named numeric vector with the entries "rss" (VmRSS) and "peak" (VmHWM), both in kB
```
pascal-sauer, line 21:
```suggestion
#' @return A named numeric vector with the entries
#' "rss" (VmRSS - virtual memory resident set size) and
#' "peak" (VmHWM - virtual memory high water mark), both in kB
```

R/memoryProfiling.R
```diff
@@ -0,0 +1,74 @@
...
+#' Checks whether the kernel interfaces required for memory profiling are available.
+#' Only Linux exposes both the high-water mark of the resident set size and a way
+#' to reset it.
+#'
+#' @return TRUE if memory profiling can be performed, FALSE otherwise
+#' @author Patrick Rein
+#' @seealso \code{\link{setConfig}}
+#' @keywords internal
+memoryProfilingSupported <- function() {
+  return(file.exists("/proc/self/status") && # nolint: absolute_path_linter.
```
pascal-sauer, line 12: sidenote: need to add these paths to the nono profile - could you do that?

R/memoryProfiling.R
```diff
@@ -0,0 +1,74 @@
...
+  memory <- as.numeric(sub("^\\D*([0-9]+).*$", "\\1", fields))
+  names(memory) <- sub("^Vm(RSS|HWM):.*$", "\\1", fields)
+  return(c(rss = memory[["RSS"]], peak = memory[["HWM"]]))
+}
+
+#' startMemoryProfiling
+#'
+#' Starts a memory profiling block by resetting the kernel's peak resident set size
+#' counter, so that the peak reported by \code{\link{reportMemoryProfiling}} refers to
+#' this block only. As that counter is process wide, profiling blocks must not overlap.
```
pascal-sauer, line 36: So if we ever get parallel madrat, then memory profiling will show nonsense, yes?

pascal-sauer (approved):

R/memoryProfiling.R
codeZeilen: Yes and no :)

`/proc` is per process so that should be fine but the data for the whole pre-processing running combined is naturally not available then. But the logging in general will be more "fun" when doing things in parallel....

R/initializeConfig.R
codeZeilen: I can only estimate, but my impression was that this is not too bad. It only runs for top-level calc calls anyways and then only reads files and does a little bit of parsing. The major downside of this would be the cluttered log.

Also we could not simply replace `FALSE` with `TRUE` here but would need to set it to `memoryProfilingSupported()` to not try to activate it on Windows.

R/memoryProfiling.R
codeZeilen: Good point, will do :)

codeZeilen: Done, but only for `/prof/self/clear_refs`, as read is given to all of `/proc` by the default linux policy already (according to `nono why`).

R/initializeConfig.R
codeZeilen: What about setting this by default in our HPC `.Rprofile`?

pascal-sauer: Sounds good

codeZeilen: On it

codeZeilen: Well, this is not that easy it turns out. Doing this in Rprofile.site leads to funny things as this loads madrat already before and renv package loading can happen....

The right way for this would be to add an environment variable. I guess I shouldn't have merged yet :P I will open a small extra PR that adds the env variable option.

PR #319, opened 2026-08-20
include mapping configuration in madrat graph hash

bodirsky:
getMadratGraph() caches the dependency graph in the in-session MadratCache, keyed by digest(mtimes, packages, globalenv). The graph's mappings attribute, however, is resolved by .getMappingFiles() against the mapping configuration active when it was built. Changing regionmapping within one session therefore reuses a graph carrying stale mapping paths; fingerprint() hashes those, and every dependent calculation resolves to the previous mapping's cache entry.

Observed in MAgPIE preprocessing: a run that computed clusters under regionmappingH12.csv and then switched to regionmappingFSEC.csv in the same session silently reused the H12 cluster result (calcCluster-F09b8144c-…) and wrote it out under the FSEC name, producing a cellular dataset whose clustermap contained 12 H12 regions instead of the 14 FSEC ones. Identical scenarios run in separate processes correctly produced calcCluster-F27e7b4af-…. There is no error — the output looks valid.

This adds regionmapping, extramappings and mappingfolder to the graph hash so a mapping switch invalidates the cached graph. Only paths are hashed; mapping contents are already covered by fingerprint() via fingerprintFiles(). Cost is one graph rebuild per mapping switch; unchanged workflows are unaffected.

Related: 56d45cbf ("preparing to include mapping files in cache hashing (incomplete)").

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '69912704'
+ValidationKey: '69977355'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '70604289'
+ValidationKey: '70628400'
 AcceptedWarnings:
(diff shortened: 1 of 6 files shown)


tscheypidi (approved):
fix merge conflicts. After that you can merge.

PR #325, opened 2026-09-04
Error on setting a cache format that is not installed

codeZeilen:


R/setConfig.R
```diff
@@ -189,7 +189,7 @@ setConfig <- function(..., # nolint: cyclocomp_linter.
   }
 
   if (!is.null(cacheformat) && .cfgchecks) {
-    cacheFormat(cacheformat)
+    checkCacheFormatAvailable(cacheformat)
```
pascal-sauer, line 192: This also sets the cache format, right? The function name implies it only checks availability

codeZeilen: It doesn't actually. The previous usage of `cacheFormat` was not ideal. `cacheFormat` returns the cacheFormat with that name. It was only there to catch a non-existent cache format. `checkCacheFormatAvailable` covers that now explicitly.

The actual cache format is stored in the `cfg` object.

pascal-sauer (approved):

PR #326, opened 2026-09-11
Fix get mapping in new cache graph

codeZeilen:
The inclusion of mappings in the cache graph causes a warning, as it revealed some undefined /  not ideal defined behavior in `toolGetMapping`. The problematic case is when `toolGetMapping` is called with an absolute path to a mapping (e.g. set via the madrat config) and a `where`. In that case the mapping path will get resolved a second time and a broken mapping path results.

This PR changes toolGetMapping to return early when getting an absolute path.


R/toolGetMapping.R
```diff
@@ -50,6 +50,11 @@ toolGetMapping <- function(name, type = NULL, where = NULL,
 }
 
 .searchName <- function(name, type, where, activecalc) {
+  # a name which already is a path to an existing file (e.g. getConfig("regionmapping")
+  # after setConfig normalized it) must not be resolved against a folder again
+  if (.isExistingFilePath(name)) {
+    return(name)
+  }
```
pascal-sauer, line 57: Can we run into trouble with this if a filename is passed that is randomly present in the current working directory?

R/toolGetMapping.R
```diff
@@ -83,6 +88,12 @@ toolGetMapping <- function(name, type = NULL, where = NULL,
   }
 }
 
+.isExistingFilePath <- function(name) {
+  name <- as.character(name)
+  return(length(name) == 1 && basename(name) != name &&
```
pascal-sauer: Ah, so we're making sure it is actually an absolute path here - can this be reflected in the function name? `.isExistingAbsolutePath`?

pascal-sauer (approved):

R/toolGetMapping.R
codeZeilen: Refactored into two properly named predicates :)

